PR #5, opened 2018-09-24
add scheduler manifest

mfojtik:
This belongs to scheduler operator, but we can't wait for it now and we need to pass the bootstrapping. Logically this is closer to controller-manager than API server.

The scheduler have hardcoded config path, @sttts I guess we will need some config that will work :)

Also adds `--skip-scheduler` flag to render (in case we want to switch this off, like when scheduler operator will provide its bootstrap manifest)

### DIFF
--- a/cmd/cluster-kube-controller-manager-operator/render/render.go
+++ b/cmd/cluster-kube-controller-manager-operator/render/render.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 
 	"github.com/ghodss/yaml"
@@ -39,6 +40,8 @@ type renderOpts struct {
 	assetOutputDir     string
 	configOverrideFile string
 	configOutputFile   string
+
+	skipSchedulerBootstrapManifest bool
 }
 
 func NewRenderCommand() *cobra.Command {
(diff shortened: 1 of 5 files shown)


cmd/cluster-kube-controller-manager-operator/render/render.go
```diff
@@ -77,6 +80,7 @@ func NewRenderCommand() *cobra.Command {
 	cmd.Flags().StringVar(&renderOpts.configOverrideFile, "config-override-file", "", "A sparse KubeControllerManagerConfig.kubecontrolplane."+
 		"config.openshift.io/v1 file (default: kube-controller-manager-config-overrides.yaml in the asset-input-dir)")
 	cmd.Flags().StringVar(&renderOpts.configOutputFile, "config-output-file", "", "Output path for the KubeControllerManagerConfig yaml file.")
+	cmd.Flags().BoolVar(&renderOpts.skipSchedulerBootstrapManifest, "skip-scheduler", false, "Skip copying the scheduler manifests.")
```
sttts: TODO: remove when a kube-scheduler operator exists

manifests/bootkube/bootstrap-manifests/kube-scheduler-pod.yaml
```diff
@@ -0,0 +1,28 @@
...
+    image: {{ .Image }}
+    imagePullPolicy: {{ .ImagePullPolicy }}
+    command: ["/bin/bash", "-c"]
+    args:
+    - exec hyperkube scheduler --config=/etc/kubernetes/config/scheduler.yaml
+    volumeMounts:
+    - mountPath: /etc/kubernetes/secrets
+      name: secrets
+      readOnly: true
+  hostNetwork: true
```
sttts: does this work? It uses the in-cluster config. I would expect that we also pass the bootstrap kubeconfig.

mfojtik: we will have to hardcode the path to it

mfojtik: added `--kubeconfig=/etc/kubernetes/secrets/kubeconfig`

aveshagarwal:
> This belongs to scheduler operator, but we can't wait for it now and we need to pass the bootstrapping. 

I have created this initial repo for cluster-kube-scheduler-operator based on existing apiserver and controller manager operators: https://github.com/aveshagarwal/cluster-kube-scheduler-operator 

I am still modifying it and not yet tested yet. 

Once I have openshift/cluster-kube-scheduler-operator created and as soon as the code starts working, I will move the code here.

So if you have anything that is related to cluster-kube-scheduler-operator, may be we could update here: https://github.com/aveshagarwal/cluster-kube-scheduler-operator

sttts:
/lgtm

Config comes in a follow-up.

PR #6, opened 2018-09-24
WIP: DO_NO_MERGE: test update deps

mfojtik:


mfojtik:
@tnozicka an indication (even a flaky one) that vendoring is broken is a good one... it will probably deserves something better long them (i dream about a dashboard that will show you what levels of library-go/etc are distributed where, which are out-of-sync and what is now broken).

PR #7, opened 2018-09-25
render: complete config

sttts:


sttts:
/hold

until smoke tests are green.

manifests/bootkube/config/bootstrap-config-overrides.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - "/etc/kubernetes/cloud/config"
+  root-ca-file:
+  - "/etc/kubernetes/secrets/root-ca.crt"
+  service-account-private-key-file:
+  - "/etc/kubernetes/secrets/service-account.key"
+  cluster-signing-cert-file:
+  - "/etc/kubernetes/secrets/kube-ca.crt"
+  cluster-signing-key-file:
+  - "/etc/kubernetes/secrets/kube-ca.key"
+  kubeconfig:
```
deads2k: turns out that for some reason we have to pass this on the command line in the pod for now.  don't know why.  Probably loading order.

sttts: good to know.

manifests/bootkube/config/config-overrides.yaml
```diff
@@ -0,0 +1,17 @@
...
+  certFile: service-signer.crt
+extendedArguments:
+  cloud-config:
+  - "/etc/kubernetes/cloud/config"
+  root-ca-file:
+  - "/etc/kubernetes/secrets/root-ca.crt"
+  service-account-private-key-file:
+  - "/etc/kubernetes/secrets/service-account.key"
+  cluster-signing-cert-file:
+  - "/etc/kubernetes/secrets/kube-ca.crt"
```
deads2k: in some followup, I think we should split secrets like we did in the apiserver

manifests/v3.11.0/kube-apiserver/defaultconfig.yaml
```diff
@@ -1,2 +1,39 @@
...
+  - "10.3.0.0/16"
+  use-service-account-credentials:
+  - "true"
+  leader-elect:
+  - "true"
+  leader-elect-retry-period:
+  - "3s"
+  leader-elect-resource-lock:
+  - "configmaps"
+  controllers:
```
deads2k: followup to remove this.  We should try to run straight upstream.

deads2k:
if it runs, I'd merge it.

manifests/bootkube/config/config-overrides.yaml
sttts: rule of thumb right now: follow kube-core to make replacing it easy. After we have done that we can make everything pretty.

sttts:
/hold

File missing.

sttts:
/hold

Manifest must not have conflicting names.

PR #10, opened 2018-09-26
make render produce minimal running config for later customization

deads2k:
We need to simplify the rendered config to be compatible with non-old cluster-up clusters.

/assign @mfojtik


bindata/v3.11.0/kube-apiserver/defaultconfig.yaml
```diff
@@ -1,14 +1,10 @@
...
 kind: KubeControllerManagerConfig
-serviceServingCert:
-  certFile: service-signer.crt
 extendedArguments:
   enable-dynamic-provisioning:
   - "true"
   allocate-node-cidrs:
   - "true"
-  cloud-provider:
-  - "metal" # TODO: override this through the installer
```
sttts: this was a placeholder. kube-core is more clever here. Do we need more?

PR #12, opened 2018-09-26
don't use hostnetwork in daemonset

deads2k:
This uses a deprecated flag (we'll need to fix that) to override the server being used by the daemonset controller manager.

/assign @mfojtik

### DIFF
--- a/bindata/bootkube/manifests/kube-controller-manager-daemonset.yaml
+++ b/bindata/bootkube/manifests/kube-controller-manager-daemonset.yaml
@@ -31,7 +31,7 @@ spec:
         imagePullPolicy: {{ .ImagePullPolicy }}
         command: ["/bin/bash", "-c"]
         args:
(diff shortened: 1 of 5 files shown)


mfojtik:
@deads2k i think we need the hostNetwork for daemonsets :)

bindata/bootkube/manifests/kube-controller-manager-daemonset.yaml
```diff
@@ -31,7 +31,7 @@ spec:
         imagePullPolicy: {{ .ImagePullPolicy }}
         command: ["/bin/bash", "-c"]
         args:
-        - exec hyperkube kube-controller-manager --openshift-config=/etc/kubernetes/config/{{ .ConfigFileName }} --kubeconfig=/etc/kubernetes/secrets/kubeconfig
+        - exec hyperkube kube-controller-manager --openshift-config=/etc/kubernetes/config/{{ .ConfigFileName }} --kubeconfig=/etc/kubernetes/secrets/kubeconfig --master=https://kubernetes.default.svc
```
sttts: @mfojtik you remember us talking about networking being available pre-operators. This DNS name certainly needs networking.

PR #19, opened 2018-10-02
add locking to prevent host port conflicts

mfojtik:
This add locking mechanism similar to what API server using to prevent the host port conflicts. For cluster up the controller manager has to use the hostNetwork otherwise the connection to API server will fail and the pods created by daemonsets will keep hanging. With this I can get the controller manager operating.

/cc @deads2k 
/cc @sttts 
/cc @openshift/sig-master


sttts:
Let's just disable the port in the bootstrap pods or use different ports. `flock` here is ugly and unnecessary. Nobody queries the phase 1 pods' port.

PR #24, opened 2018-10-02
tolerate any taint

deads2k:
/assign @abhinavdahiya @mfojtik

### DIFF
--- a/bindata/v3.11.0/kube-apiserver/deployment.yaml
+++ b/bindata/v3.11.0/kube-apiserver/deployment.yaml
@@ -84,3 +84,5 @@ spec:
           secretName: serving-cert
       nodeSelector:
         node-role.kubernetes.io/master: ""
+      tolerations:
+      - operator: Exists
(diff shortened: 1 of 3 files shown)


bindata/v3.11.0/kube-apiserver/deployment.yaml
```diff
@@ -84,3 +84,5 @@ spec:
           secretName: serving-cert
```
abhinavdahiya: Why is this file called `kube-apiserver/deployment.yaml` when its controller-manager ?

abhinavdahiya:
`/lgtm`

The tolerate all is from https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/#concepts

bindata/v3.11.0/kube-apiserver/deployment.yaml
deads2k: > Why is this file called `kube-apiserver/deployment.yaml` when its controller-manager ?

It's content is a deployment

abhinavdahiya: `deployment.yaml` seems fine, just that the directory was `kube-apiserver`?
Its an trivial curiosity.

PR #28, opened 2018-10-03
Ensure operator config exists and has the right image when operator start

mfojtik:
@deads2k i think this is what you meant and we want in all operators? ;-)

(second commit)

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -2,13 +2,21 @@ package operator
 
 import (
 	"fmt"
+	"os"
 	"time"
 
+	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/apis/kubecontrollermanager/v1alpha1"
+	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/v311_00_assets"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
 	operatorconfigclient "github.com/openshift/cluster-kube-controller-manager-operator/pkg/generated/clientset/versioned"
+	operatorsv1alpha1client "github.com/openshift/cluster-kube-controller-manager-operator/pkg/generated/clientset/versioned/typed/kubecontrollermanager/v1alpha1"
 	operatorclientinformers "github.com/openshift/cluster-kube-controller-manager-operator/pkg/generated/informers/externalversions"
 )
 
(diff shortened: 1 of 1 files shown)


mfojtik:
/hold 

for @deads2k review/comments

pkg/operator/sync_kubecontrollermanager_v311_00.go
```diff
@@ -24,6 +30,9 @@ func syncKubeControllerManager_v311_00_to_latest(c KubeControllerManagerOperator
 	versionAvailability := operatorsv1alpha1.VersionAvailablity{
 		Version: operatorConfig.Spec.Version,
 	}
+	v1alpha1Scheme := runtime.NewScheme()
```
deads2k: check to see if it exists before you do any of this.

pkg/operator/sync_kubecontrollermanager_v311_00.go
```diff
@@ -58,12 +67,34 @@ func syncKubeControllerManager_v311_00_to_latest(c KubeControllerManagerOperator
 		errors = append(errors, fmt.Errorf("%q: %v", "sa", err))
 	}
 
+	operatorConfigBytes := v311_00_assets.MustAsset("v3.11.0/kube-controller-manager/operator-config.yaml")
+	operatorConfigObj, err := runtime.Decode(v1alpha1Codecs.UniversalDecoder(v1alpha1.GroupVersion), operatorConfigBytes)
```
deads2k: just panic if it errors, that's a coding error

pkg/operator/sync_kubecontrollermanager_v311_00.go
```diff
@@ -58,12 +67,34 @@ func syncKubeControllerManager_v311_00_to_latest(c KubeControllerManagerOperator
...
+	if err != nil {
+		errors = append(errors, err)
+	}
+	requiredOperatorConfig, ok := operatorConfigObj.(*v1alpha1.KubeControllerManagerOperatorConfig)
+	if !ok {
+		errors = append(errors, fmt.Errorf("operator-config.yaml is not v1alpha1.KubeControllerManagerOperatorConfig (%t)", operatorConfigObj))
```
deads2k: panic

pkg/operator/sync_kubecontrollermanager_v311_00.go
```diff
@@ -58,12 +67,34 @@ func syncKubeControllerManager_v311_00_to_latest(c KubeControllerManagerOperator
...
+	}
+	imagePullSpec := os.Getenv("IMAGE")
+	if len(imagePullSpec) == 0 {
+		errors = append(errors, fmt.Errorf("required environment variable IMAGE is not set"))
```
deads2k: don't fail here, just don't overwrite the value if it isn't present and we can fill in a default value in our bindata

pkg/operator/sync_kubecontrollermanager_v311_00.go
```diff
@@ -58,12 +67,34 @@ func syncKubeControllerManager_v311_00_to_latest(c KubeControllerManagerOperator
...
+		errors = append(errors, err)
+	}
+
 	controllerManagerConfig, configMapModified, err := manageKubeControllerManagerConfigMap_v311_00_to_latest(c.corev1Client, operatorConfig)
 	if err != nil {
 		errors = append(errors, fmt.Errorf("%q: %v", "configmap", err))
 	}
 
 	forceDeployment := operatorConfig.ObjectMeta.Generation != operatorConfig.Status.ObservedGeneration
+	if configModified {
```
deads2k: this isn't needed.

pkg/operator/sync_kubecontrollermanager_v311_00.go
```diff
@@ -90,6 +121,28 @@ func syncKubeControllerManager_v311_00_to_latest(c KubeControllerManagerOperator
 	return resourcemerge.ApplyGenerationAvailability(versionAvailability, actualDeployment, errors...), errors
 }
 
+func applyKubeControllerManagerOperatorConfig(client operatorsv1alpha1client.KubeControllerManagerOperatorConfigsGetter, required *v1alpha1.KubeControllerManagerOperatorConfig) (*v1alpha1.KubeControllerManagerOperatorConfig, bool, error) {
```
deads2k: this function needs to be weird.  Only adjust the imagepullspec if it mismatches or create if it is missing entirely.

pkg/operator/sync_kubecontrollermanager_v311_00.go
deads2k: move all the logic to be done synchronously in the controller's run method.

pkg/operator/starter.go
```diff
@@ -34,9 +44,56 @@ func RunOperator(clientConfig *rest.Config, stopCh <-chan struct{}) error {
...
+	if err != nil {
+		panic(err)
+	}
+	requiredOperatorConfig, ok := operatorConfigObj.(*v1alpha1.KubeControllerManagerOperatorConfig)
+	if !ok {
+		panic(fmt.Sprintf("unexpected object in %s: %t", path, operatorConfigObj))
+	}
+
+	imagePullSpec := os.Getenv("IMAGE")
+	if len(imagePullSpec) == 0 {
```
deads2k: invert this check and only change the value if the env var is set.

pkg/operator/starter.go
```diff
@@ -34,9 +44,56 @@ func RunOperator(clientConfig *rest.Config, stopCh <-chan struct{}) error {
...
+	if errors.IsNotFound(err) {
+		if _, err := client.KubeControllerManagerOperatorConfigs().Create(requiredOperatorConfig); err != nil {
+			panic(err)
+		}
+		return
+	}
+	if err != nil {
+		panic(err)
+	}
+
```
deads2k: if the env var wasn't set, return early here, because we shouldn't change the value.

pkg/operator/starter.go
```diff
@@ -34,9 +44,56 @@ func RunOperator(clientConfig *rest.Config, stopCh <-chan struct{}) error {
...
+	modified := resourcemerge.BoolPtr(false)
+	resourcemerge.EnsureObjectMeta(modified, &existing.ObjectMeta, requiredOperatorConfig.ObjectMeta)
+	dataSame := equality.Semantic.DeepEqual(existing.Spec, requiredOperatorConfig.Spec)
+	if dataSame && !*modified {
+		return
+	}
+	existing.Spec = requiredOperatorConfig.Spec
```
deads2k: only set the image pull sec.

pkg/operator/starter.go
```diff
@@ -34,9 +44,56 @@ func RunOperator(clientConfig *rest.Config, stopCh <-chan struct{}) error {
...
+			panic(err)
+		}
+		return
+	}
+	if err != nil {
+		panic(err)
+	}
+
+	modified := resourcemerge.BoolPtr(false)
+	resourcemerge.EnsureObjectMeta(modified, &existing.ObjectMeta, requiredOperatorConfig.ObjectMeta)
```
deads2k: don't bother with this check

PR #36, opened 2018-10-10
add image references

juanvallejo:
cc @deads2k


manifests/0000_12_cluster-kube-controller-manager-operator_06_deployment.yaml
```diff
@@ -28,6 +28,9 @@ spec:
         volumeMounts:
         - mountPath: /var/run/configmaps/config
           name: config
+          env:
```
deads2k: this looks like it is nested too far.

PR #42, opened 2018-10-12
Makefile: add origin-release target

sttts:


Makefile
```diff
@@ -1,4 +1,5 @@
 GOFLAGS :=
+DOCKER_ORG ?= openshift
```
soltysh: repo org, user org, not docker please.

hack/lib/Dockerfile-origin-release
```diff
@@ -0,0 +1,9 @@
+FROM registry.svc.ci.openshift.org/openshift/origin-release:v4.0 as origin-release
+
+FROM fedora:29 as jq
```
soltysh: Fedora and not centos?

soltysh: What does `as jq` line does here?

mfojtik: not sure if `jq` is avaiable in centos:7, however I would probably just use some minimal jq (scratch image exists?).

mfojtik: or will just instead of yum install curl it directly on centos:7 so I don't have to pull the fedora layers ;-)

soltysh: There's none by default, you need to install `epel-release` and only then `jq`. Or just `wget` it.

PR #47, opened 2018-10-16
make fake checkpointer more real

mfojtik:
Copied from https://github.com/coreos-inc/tectonic-operators/blob/master/operator/kube-core/spec/manifests/pod-checkpointer.yaml

I think we need real checkpointer to pass the bootkube process. Placement in this repository is just temporary until we find the proper owner (or have checkpointer operator).

/cc @derekwaynecarr


sttts:
lgtm

if smoke test is green.

mfojtik:
Lets do this in one place: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/47

PR #48, opened 2018-10-16
Fix several bootstrap problems for controller-manager

mfojtik:
This PR will:

* Add temporary kube-proxy manifests (+role/etc)
* Add temporary but more real checkpointer (+role/etc)

These two additions should be immediately removed after we lift the control plane off ground.
Additionally this PR will

* Update daemonset to use hostNetwork (might not be needed as with kube-proxy the tectonic network operator should run in-cluster network)
* Fix wrong command and wrong flags (--config vs. --openshift-config)


bindata/bootkube/manifests/daemonset-kube-controller-manager.yaml
```diff
@@ -43,14 +42,15 @@ spec:
...
-        - mountPath: /var/run/secrets/client-ca
+        - mountPath: /var/run/configmaps/client-ca
           name: client-ca
       nodeSelector:
         node-role.kubernetes.io/master: ""
       tolerations:
       - key: node-role.kubernetes.io/master
         operator: Exists
         effect: NoSchedule
+      hostNetwork: true
```
mfojtik: @sttts this might not be needed after all if the networking operator is able to reach the API server after we deploy the kube-proxy

sttts:
Replaced by https://github.com/openshift/cluster-kube-controller-manager-operator/pull/49.

bindata/bootkube/manifests/daemonset-kube-controller-manager.yaml
sttts: Probably the phase-1 controller manager creates all necessary service accounts and what not. Then this comes up even without host networking. Have the slight feeling that this might work most often, but it's more a conincidence.

PR #50, opened 2018-10-16
rename conflicting file

deads2k:
when two render steps create files with the same name, only one wins during a copy and we got the "wrong" one.


deads2k:
@mfojtik @sttts why didn't I have a bindata update?

deads2k:
@sjenning thanks for the issue.  This and the other should fix it.

deads2k:
lgtm'd via slack.

PR #52, opened 2018-10-17
Fix deployment command name

sttts:


bindata/v3.11.0/kube-controller-manager/deployment.yaml
```diff
@@ -25,7 +25,7 @@ spec:
       - name: controller-manager
         image: ${IMAGE}
         imagePullPolicy: IfNotPresent
-        command: ["hypershift", "openshift-kube-controller-manager"]
+        command: ["hyperkube", "kube-controller-manager"]
         args:
         - "--config=/var/run/configmaps/config/config.yaml"
```
mfojtik: --openshift-config

sttts: fixed

PR #60, opened 2018-10-18
switch to daemonsets

deads2k:
Switch the kube-controller-manager to a daemonset and delete the old bootkube daemonset once we're up so we are sure to take the load and actually work

/assign @damemi @mfojtik


pkg/operator/sync_kubecontrollermanager_v311_00.go
```diff
@@ -28,66 +27,60 @@ func syncKubeControllerManager_v311_00_to_latest(c KubeControllerManagerOperator
...
+	actualDaemonSet, _, err := manageKubeControllerManagerDaemonSet_v311_00_to_latest(c.kubeClient.AppsV1(), operatorConfig, previousAvailability, forceRollout)
 	if err != nil {
-		errors = append(errors, fmt.Errorf("%q: %v", "deployment", err))
+		errors = append(errors, fmt.Errorf("%q: %v", "daemonset", err))
+	}
+
+	// if our actual daemonset has at least one pod available, then we should delete the kube-system daemonset to make sure we take over for it.
+	if actualDaemonSet.Status.NumberAvailable > 0 {
+		// we don't care about the return value here.
+		c.kubeClient.AppsV1().DaemonSets("kube-system").Delete("kube-controller-manager", nil)
```
deads2k: @sttts I really want this delete in order to be sure that our controller manager actually works (it didn't before).  Is it going to break bootkube?

bindata/v3.11.0/kube-controller-manager/clusterrolebinding.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: openshift-kube-controller-manager-cluster-admin
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
```
deads2k: the kubeconfig for the "other" daemonset was getting a `system:masters` identity, so this is equivalent for now.

pkg/operator/sync_kubecontrollermanager_v311_00.go
sttts: We move the phase 2 daemonset into the operator namespace. Then we can render exactly the manifest you would create in phase 3. So you can easily adopt it.

bindata/v3.11.0/kube-controller-manager/clusterrolebinding.yaml
sttts: Should we create an issue that this is to be fixed?

deads2k: > Should we create an issue that this is to be fixed?

Sure, but I wouldn't start working on it now

deads2k:
we are moving in a different direction.

PR #65, opened 2018-10-30
Wire cloud-provider to kube-controller-manager

ironcladlou:
Fixes https://github.com/openshift/installer/issues/577.

Related to https://github.com/openshift/installer/issues/530.

### DIFF
--- a/pkg/operator/observe_config.go
+++ b/pkg/operator/observe_config.go
@@ -7,8 +7,10 @@ import (
 	"reflect"
 	"time"
 
+	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -44,6 +46,7 @@ type ConfigObserver struct {
 func NewConfigObserver(
 	operatorConfigInformer operatorconfiginformerv1alpha1.KubeControllerManagerOperatorConfigInformer,
 	kubeInformersForOpenShiftKubeControllerManagerNamespace informers.SharedInformerFactory,
+	kubeInformersForKubeSystemNamespace informers.SharedInformerFactory,
 	operatorConfigClient operatorconfigclientv1alpha1.KubecontrollermanagerV1alpha1Interface,
 	kubeClient kubernetes.Interface,
 	clientConfig *rest.Config,
@@ -56,11 +59,14 @@ func NewConfigObserver(
 		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "ConfigObserver"),
 
 		rateLimiter: flowcontrol.NewTokenBucketRateLimiter(0.05 /*3 per minute*/, 4),
-		observers:   []observeConfigFunc{},
+		observers: []observeConfigFunc{
+			observeClusterConfig,
+		},
 	}
 
 	operatorConfigInformer.Informer().AddEventHandler(c.eventHandler())
 	kubeInformersForOpenShiftKubeControllerManagerNamespace.Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
+	kubeInformersForKubeSystemNamespace.Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
 	return c
 }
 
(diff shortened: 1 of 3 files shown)


pkg/operator/observe_config_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/rest"
+)
+
+func TestObserveClusterConfig(t *testing.T) {
```
ironcladlou: Test needs refactored to use tables, not going to bother until we're consuming data from the right place

pkg/operator/observe_config.go
```diff
@@ -97,6 +101,58 @@ func (c ConfigObserver) sync() error {
...
+	if err != nil {
+		glog.Warningf("Unable to parse install-config: %s", err)
+		return observedConfig, nil
+	}
+
+	// extract needed values
+	//  data:
+	//   install-config:
+	//     platform:
+	//       aws: {}
```
ironcladlou: This isn't the right place to get cloud provider info from; I'm just using it for testing.

pkg/operator/observe_config.go
```diff
@@ -97,6 +101,58 @@ func (c ConfigObserver) sync() error {
...
+	if errors.IsNotFound(err) {
+		glog.Warningf("cluster-config-v1 not found in the kube-system namespace")
+		return observedConfig, nil
+	}
+	if err != nil {
+		return observedConfig, err
+	}
+
+	installConfigYaml, ok := clusterConfig.Data["install-config"]
+	if !ok {
```
juanvallejo: check for the length here, in case of an empty string

pkg/operator/observe_config.go
```diff
@@ -97,6 +101,58 @@ func (c ConfigObserver) sync() error {
...
+	}
+
+	// extract needed values
+	//  data:
+	//   install-config:
+	//     platform:
+	//       aws: {}
+	cloudProvider := ""
+	platform, ok := installConfig["platform"].(map[string]interface{})
+	if !ok {
```
juanvallejo: check for empty values as well

pkg/operator/observe_config.go
```diff
@@ -97,6 +101,58 @@ func (c ConfigObserver) sync() error {
...
+	}
+	if err != nil {
+		return observedConfig, err
+	}
+
+	installConfigYaml, ok := clusterConfig.Data["install-config"]
+	if !ok {
+		return observedConfig, nil
+	}
+	installConfig := map[string]interface{}{}
```
abhinavdahiya: would like to see if we can vendor installer's type rather than this map[string]interface{}, install-config has Name func on platform [see here](https://github.com/openshift/installer/blob/master/pkg/types/installconfig.go#L74-L91) that can be used to make decisions.

deads2k: > would like to see if we can vendor installer's type rather than this map[string]interface{}, install-config has Name func on platform [see here](https://github.com/openshift/installer/blob/master/pkg/types/installconfig.go#L74-L91) that can be used to make decisions.

Once the type is promoted to `openshift/api` we can do that, but I don't want to start having different operators depend on each other for types.  If you do that, we'll end up wedged as we try to rebase to various levels of kube.

pkg/operator/observe_config.go
```diff
@@ -97,6 +101,58 @@ func (c ConfigObserver) sync() error {
 	return nil
 }
 
+// observeClusterConfig observes cloud provider configuration from
+// cluster-config-v1 in order to configure kube-controller-manager's cloud
+// provider.
+func observeClusterConfig(kubeClient kubernetes.Interface, clientConfig *rest.Config, observedConfig map[string]interface{}) (map[string]interface{}, error) {
+	clusterConfig, err := kubeClient.CoreV1().ConfigMaps("kube-system").Get("cluster-config-v1", metav1.GetOptions{})
```
deads2k: I don't see the namespace scoped informer being plumbed to the control loop to watch for events.

pkg/operator/observe_config.go
```diff
@@ -97,6 +101,58 @@ func (c ConfigObserver) sync() error {
...
+	//       aws: {}
+	cloudProvider := ""
+	platform, ok := installConfig["platform"].(map[string]interface{})
+	if !ok {
+		return observedConfig, nil
+	}
+	if aws := platform["aws"]; aws != nil {
+		cloudProvider = "aws"
+	} else {
+		glog.Warningf("No recognized cloud provider found in install-config/platform")
```
deads2k: if we expect more, make it a switch case.

Also, why can't we return here?

pkg/operator/observe_config.go
```diff
@@ -97,6 +101,58 @@ func (c ConfigObserver) sync() error {
...
+	}
+
+	// set observed values
+	//  extendedArguments:
+	//    cloud-provider:
+	//	  - "name"
+	if len(cloudProvider) > 0 {
```
deads2k: if we returned above, is this `if` still needed or can we do it unconditionally?

pkg/operator/observe_config.go
ironcladlou: Fixed (I think)

pkg/operator/observe_config.go
ironcladlou: Refactored

pkg/operator/observe_config.go
ironcladlou: Refactored

pkg/operator/observe_config.go
ironcladlou: Does it matter given we check for unmarshaling errors?

ironcladlou: Oops, previous comment applies to another line... what do you mean re: empty value here? if the type conversion works I have a map to work with

pkg/operator/observe_config.go
ironcladlou: Not sure it matters given we check for deserialization errors

pkg/operator/observe_config.go
ironcladlou: Yeah, this is a pervasive issue that cuts across all my team's operators (and now this one). Not going to solve it with this PR.

ironcladlou:
Just to be totally clear, the current PR is a WIP; I'm inferring cloud provider config from some existing cluster config that happens to be useful but which isn't really designed to be cloud provider config. Before this is mergeable we might need an installer PR to introduce cloud provider stuff to the cluster config and then refactor this code to use the new purposefully introduced cloud provider config.

pkg/operator/observe_config.go
deads2k: > Not sure it matters given we check for deserialization errors

In preference to an `ok`.  That's fairly standard in the overall kube codebase.

ironcladlou:
Squashed

pkg/operator/observe_config.go
```diff
@@ -97,6 +103,60 @@ func (c ConfigObserver) sync() error {
...
+// cluster-config-v1 in order to configure kube-controller-manager's cloud
+// provider.
+func observeClusterConfig(kubeClient kubernetes.Interface, clientConfig *rest.Config, observedConfig map[string]interface{}) (map[string]interface{}, error) {
+	clusterConfig, err := kubeClient.CoreV1().ConfigMaps("kube-system").Get("cluster-config-v1", metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		glog.Warningf("cluster-config-v1 not found in the kube-system namespace")
+		return observedConfig, nil
+	}
+	if err != nil {
+		return observedConfig, err
```
juanvallejo: nit: in the future, might be good to just explicitly return `nil` here for observedConfig

pkg/operator/observe_config.go
```diff
@@ -97,6 +103,60 @@ func (c ConfigObserver) sync() error {
...
+
+	installConfigYaml, ok := clusterConfig.Data["install-config"]
+	if !ok {
+		return observedConfig, nil
+	}
+	installConfig := map[string]interface{}{}
+	err = yaml.Unmarshal([]byte(installConfigYaml), &installConfig)
+	if err != nil {
+		glog.Warningf("Unable to parse install-config: %s", err)
+		return observedConfig, nil
```
juanvallejo: any particular reason why we would not return the error here?

ironcladlou:
e2e failure is reminiscent of https://github.com/openshift/installer/pull/606

ironcladlou:
This is what I get for rebasing. 😭

/retest

ironcladlou:
From my testing I'm pretty sure this also fixed https://github.com/openshift/installer/issues/530.

PR #73, opened 2018-11-02
phase-3: add missing certs+keys

sttts:
We forgot the serviceaccount key and the cert signer key+cert in phase 3.


sttts:
Manual tests are green. This should fix the ingress to start up (failed before due to missing service account token).

PR #77, opened 2018-11-09
add liveness and readiness probes to static pod

mfojtik:
@deads2k @sttts lets discuss the probes under this PR...

Right now, we have wrong port on the service and I don't think so we serving the healthz or metrics security from controller manager. Also the `10252` seems odd and we need to decide if we keep it or just move to something else? (8444 before).

### DIFF
--- a/bindata/bootkube/config/bootstrap-config-overrides.yaml
+++ b/bindata/bootkube/config/bootstrap-config-overrides.yaml
@@ -9,6 +9,8 @@ extendedArguments:
   - "/etc/kubernetes/secrets/kube-ca.crt"
   cluster-signing-key-file:
   - "/etc/kubernetes/secrets/kube-ca.key"
+  secure-port:
+  - "10257"
   port:
   - "0"
   {{if .ClusterCIDR }}
--- a/bindata/v3.11.0/kube-controller-manager/pod.yaml
+++ b/bindata/v3.11.0/kube-controller-manager/pod.yaml
@@ -21,9 +21,25 @@ spec:
       requests:
         memory: 200Mi
         cpu: 100m
+    ports:
+      - containerPort: 10257
     volumeMounts:
     - mountPath: /etc/kubernetes/static-pod-resources
       name: resource-dir
+    livenessProbe:
+      httpGet:
+        scheme: HTTPS
+        port: 10257
+        path: healthz
+      initialDelaySeconds: 45
+      timeoutSeconds: 10
+    readinessProbe:
+      httpGet:
+        scheme: HTTPS
+        port: 10257
+        path: healthz
+      initialDelaySeconds: 10
+      timeoutSeconds: 10
   hostNetwork: true
   priorityClassName: system-node-critical
   tolerations:
--- a/bindata/v3.11.0/kube-controller-manager/svc.yaml
+++ b/bindata/v3.11.0/kube-controller-manager/svc.yaml
@@ -13,4 +13,4 @@ spec:
   ports:
   - name: https
     port: 443
-    targetPort: 8443
+    targetPort: 10257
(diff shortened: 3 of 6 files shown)


bindata/v3.11.0/kube-controller-manager/pod.yaml
```diff
@@ -19,6 +19,11 @@ spec:
     volumeMounts:
     - mountPath: /etc/kubernetes/static-pod-resources
       name: resource-dir
+    livenessProbe:
+      httpGet:
+        scheme: HTTP
+        port: 10252
```
deads2k: @sttts chose a secure one

bindata/v3.11.0/kube-controller-manager/svc.yaml
```diff
@@ -4,13 +4,12 @@ metadata:
   namespace: openshift-kube-controller-manager
   name: controller-manager
   annotations:
-    service.alpha.openshift.io/serving-cert-secret-name: serving-cert
     prometheus.io/scrape: "true"
-    prometheus.io/scheme: https
+    prometheus.io/scheme: http
```
deads2k: I expected us to remain https.  Why would we choose not to?

sttts: 1.12 has https in controller manager. Did we carry patch it in 3.11?

mfojtik: we can wait for the 1.12 rebase to land maybe?

mfojtik:
/hold

Waiting for 1.12.2 rebase and controller manager that have secure healthz endpoint.

bindata/v3.11.0/kube-controller-manager/svc.yaml
sttts: 1.12 is in. We can switch back.

bindata/v3.11.0/kube-controller-manager/pod.yaml
sttts: :+1:

bindata/v3.11.0/kube-controller-manager/svc.yaml
sttts: https?

bindata/bootkube/config/bootstrap-config-overrides.yaml
```diff
@@ -9,8 +9,8 @@ extendedArguments:
   - "/etc/kubernetes/secrets/kube-ca.crt"
   cluster-signing-key-file:
   - "/etc/kubernetes/secrets/kube-ca.key"
-  port:
-  - "0"
```
mfojtik: @deads2k @sttts this is new, we were not doing liveness probe during bootstrap, I wonder if we want to start now.

bindata/v3.11.0/kube-controller-manager/pod.yaml
```diff
@@ -21,9 +21,25 @@ spec:
...
+      - containerPort: 10257
     volumeMounts:
     - mountPath: /etc/kubernetes/static-pod-resources
       name: resource-dir
+    livenessProbe:
+      httpGet:
+        scheme: HTTPS
+        port: 10257
+        path: healthz
+      initialDelaySeconds: 45
```
mfojtik: copied from kube-apiserver, i'm up for suggestion on better numbers (i think we need some initial delay till the controller manager apiserver boots up)

bindata/bootkube/config/bootstrap-config-overrides.yaml
sttts: maybe that's not wrong. Use `0` here to disable http.

Issue #80, opened 2018-11-13
Network CIDRs not observed from cluster config


sttts:
We need the counterpart of https://github.com/openshift/cluster-kube-apiserver-operator/pull/74 for the controller manager. We have it for the bootstrapping phase (merged in https://github.com/openshift/cluster-kube-controller-manager-operator/pull/76), but not post-bootstrap.

damemi:
/assign @damemi 

@sanchezl if you've already started work on this or want to take it feel free

sttts:
/close

Fixed in https://github.com/openshift/cluster-kube-controller-manager-operator/pull/84.

PR #84, opened 2018-11-14
Observe network cidrs

sanchezl:
observe network CIDRs from cluster-config-v1.kube-system:
```yaml
data:
  install-config:
    networking:
    podCIDR: 10.2.0.0/16
    serviceCIDR: 10.3.0.0/16
```
as:
```yaml
extendedArguments:
  cluster-cidr:
  - 10.2.0.0/16
  service-cluster-ip-range:
  - 10.3.0.0/16
```
Resolves issue #80


pkg/operator/observe_config.go
```diff
@@ -219,6 +221,92 @@ func observeCloudProviderNames(listers Listers, existingConfig map[string]interf
...
+	installConfig := map[string]interface{}{}
+	err = yaml.Unmarshal([]byte(installConfigYaml), &installConfig)
+	if err != nil {
+		glog.Warningf("Unable to parse install-config: %s", err)
+		return observedConfig, errs
+	}
+
+	podCIDR, _, _ := unstructured.NestedString(installConfig, "networking", "podCIDR")
+	if len(podCIDR) == 0 {
+		errs = append(errs, fmt.Errorf("configmap/cluster-config-v1.kube-system: install-config/networking/podCIDR not found"))
```
sttts: JSON paths should be separated with dots: "install-config networking.podCIDR ...."

sttts: Is `configmap/cluster-config-v1.kube-system` what we use for fully qualified resources? Would expect `configmap kube-system/cluster-config-v1`

pkg/operator/observe_config.go
```diff
@@ -219,6 +221,92 @@ func observeCloudProviderNames(listers Listers, existingConfig map[string]interf
...
+
+	installConfigYaml, ok := clusterConfig.Data["install-config"]
+	if !ok {
+		glog.Warning("configmap/cluster-config-v1.kube-system: install-config not found")
+		return observedConfig, errs
+	}
+	installConfig := map[string]interface{}{}
+	err = yaml.Unmarshal([]byte(installConfigYaml), &installConfig)
+	if err != nil {
+		glog.Warningf("Unable to parse install-config: %s", err)
```
sttts: why is this just a warning, but the case below is returned as error?

deads2k: should be error

sanchezl: :+1:

sanchezl: :+1:

PR #89, opened 2018-11-19
 use generic config observer

deads2k:
Just a code shuffle to use the standard handling and merging logic.

Looks like the methods themselves don't work though.

/assign @sanchezl


pkg/operator/configobservation/network/observe_network.go
```diff
@@ -0,0 +1,104 @@
...
+
+import (
+	"fmt"
+
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+
+	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/configobservation"
+
+	"github.com/golang/glog"
+	yaml "gopkg.in/yaml.v2"
```
sanchezl: Use `"github.com/ghodss/yaml"` instead (returns `map[string]interface{}` instead of `map[interface{}]interface{}`).

pkg/operator/configobservation/cloudprovider/observe_cloudprovider.go
```diff
@@ -0,0 +1,78 @@
+package cloudprovider
+
+import (
+	"fmt"
+
+	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/configobservation"
```
sttts: nit: strange order

deads2k: > nit: strange order

updated

PR #95, opened 2018-11-29
make static pods node critical priority

mfojtik:
@aveshagarwal asking here... if this is a static pod with nodeSelector filled in to pin it into specific master node, will the pod priority make any difference? (i don't think we use scheduler to schedule these, right?)

/cc @smarterclayton


aveshagarwal:
@mfojtik need to assign priority class to operator too.

PR #96, opened 2018-11-30
operator: validate kube-controller-manager config before starting deployment of a new revision

sttts:


deads2k:
Rather than do this, can we have a graceful shutdown that updates our lease to a super short time before death?

/hold

deads2k:
When you rebase, please target specific items in the config and provide an event and open an issue to mark us as failing or unavailable or something.

/hold cancel

sttts:
/hold

addressing https://github.com/openshift/cluster-kube-controller-manager-operator/pull/96#issuecomment-443933270 first.

pkg/operator/target_config_reconciler_v311_00.go
```diff
@@ -82,6 +85,14 @@ func createTargetConfigReconciler_v311_00_to_latest(c TargetConfigReconciler, re
 }
 
 func manageKubeControllerManagerConfigMap_v311_00_to_latest(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, operatorConfig *v1alpha1.KubeControllerManagerOperatorConfig) (*corev1.ConfigMap, bool, error) {
+	var observedConfig map[string]interface{}
```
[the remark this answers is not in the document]
mfojtik: I would keep this logic here and set the condition for the installer controller and block installer controller probably if the condition is true? (that will prevent running installation of revision we know already that will fail)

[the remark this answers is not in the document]
sttts: @deads2k need your opinion here

pkg/operator/configobservation/validation/validate.go
```diff
@@ -0,0 +1,9 @@
+package validation
+
+import "github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/configobservation/network"
+
+// Validate verifies whether the observed configuration can be used to make progress.
+func Validate(observedConfig map[string]interface{}) (errs []error) {
```
deads2k: this is correct, but extremely broad for abuse and drift.  How about a simple list of required keys?

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/revision/deployment_controller.go
```diff
@@ -106,6 +115,22 @@ func (c RevisionController) createRevisionIfNeeded(operatorSpec *operatorv1.Oper
 		return true, nil
 	}
 
+	if c.validate != nil {
+		if err := c.validate(configs, secrets); err != nil {
+			cond := operatorv1.OperatorCondition{
+				Type:    "RevisionControllerFailing",
+				Status:  operatorv1.ConditionTrue,
+				Reason:  "ContentInvalid",
+				Message: err.Error(),
```
deads2k: that helper function I described to @mfojtik here seems to apply https://github.com/openshift/library-go/pull/150#discussion_r244737541

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/revision/deployment_controller.go
```diff
@@ -106,6 +115,22 @@ func (c RevisionController) createRevisionIfNeeded(operatorSpec *operatorv1.Oper
 		return true, nil
 	}
 
+	if c.validate != nil {
+		if err := c.validate(configs, secrets); err != nil {
```
deads2k: doesn't this validation need to happen *before* we create the configmaps and secrets for the revision?

sttts: Does not matter much. We have to make sure the config synched to the revisioned variants is correct, all at once and atomically checked. We could read all configmaps and secrets at once, validate them and sync those. But we cannot (form atomicity reasons) validate by reading, then read again to sync. Hence, my change is small, maybe slightly ugly. Valdating before the sync means to change the whe sync logic. Is it worth it?

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/revision/deployment_controller.go
sttts: fixed to use `NewMultiLineAggregate`

PR #106, opened 2018-12-05
add memory request to kube-controller-manager and operator pod

sjenning:
One in a series of PRs to set memory requests for all control plane and high memory use components.

We are encountering master instability (https://github.com/openshift/installer/pull/408 https://github.com/openshift/installer/pull/785) as the number of components increase because many components run `BestEffort` giving the scheduler no information on pod resource usage.

On a steady state empty cluster, `kube-controller-manager` uses 143MB.

{pod_name="openshift-kube-controller-manager-dev-master-0"} | 142659584

Because it is a control plane component, I'm giving it a buffer and setting to `200Mi`

@deads2k @sttts  @derekwaynecarr


sjenning:
e2e-aws is touchy presently
/retest

PR #110, opened 2018-12-12
bindata/bootkube: use root-ca.crt for the root-ca-file

abhinavdahiya:
https://github.com/openshift/cluster-kube-controller-manager-operator/blob/60ddae3e99ce150f5befb34cbcfe4318577c8bd7/bindata/bootkube/config/bootstrap-config-overrides.yaml already uses `root-ca.crt` for root-ca-file
https://github.com/openshift/cluster-kube-controller-manager-operator/blob/60ddae3e99ce150f5befb34cbcfe4318577c8bd7/bindata/bootkube/manifests/configmap-kube-controller-manager-client-ca.yaml#L8 is using `kube-ca.crt` that ends up being used as root-ca-file
here https://github.com/openshift/cluster-kube-controller-manager-operator/blob/60ddae3e99ce150f5befb34cbcfe4318577c8bd7/bindata/bootkube/config/config-overrides.yaml#L5

*I am not completely sure that how the root-ca-file gets set in bootstrap and long running phase.* :innocent: 

/cc @aaronlevy @deads2k

PR #112, opened 2018-12-17
Include kube-ca in client-ca bundle

mrogers950:
https://github.com/openshift/cluster-kube-controller-manager-operator/pull/110 caused cluster monitoring to break since the kube-ca intermediate is no longer loaded. Until we can ensure that across the board server certificates are distributed with their intermediate signer, we should include it in the ca bundle.
Re: https://jira.coreos.com/browse/MON-498
/cc @s-urbaniak @openshift/sig-auth @abhinavdahiya

### DIFF
--- a/bindata/bootkube/manifests/configmap-kube-controller-manager-client-ca.yaml
+++ b/bindata/bootkube/manifests/configmap-kube-controller-manager-client-ca.yaml
@@ -6,4 +6,5 @@ metadata:
 data:
   ca-bundle.crt: |
     {{ .Assets | load "root-ca.crt" | indent 4 }}
+    {{ .Assets | load "kube-ca.crt" | indent 4 }}
 


s-urbaniak:
cc @deads2k @smarterclayton @sjenning @eparis this should fix the scraping kubelet problem.

sttts:
This ends up as `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` in pods, used to validate the apiserver's server cert.

/lgtm
/approve

sttts:
Alternatively: add the kube-ca to the apiserver cert in https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/bootkube/manifests/secret-serving-cert.yaml. Would make more sense, wouldn't it? /cc @abhinavdahiya @deads2k

PR #113, opened 2018-12-18
Revert "Include kube-ca in client-ca bundle"

sttts:
Reverts openshift/cluster-kube-controller-manager-operator#112

The kube-ca has to be served as part of the server certs. This is the case for the apiserver (otherwise, more would have been broken). For some reason the cadvisor does not serve it, which broke @s-urbaniak's use-case.

I don't think it is right to distribute the kube-ca through the whole cluster as part of some CA bundle. Being part of the server certs is the right approach instead.

Independently, it is questionably why the service account ca.crt is the right CA to be used against the cadvisor in the first place. That CA is meant for the apiserver, and just happens to match other services on the node.

@sjenning can you comment?

### DIFF
--- a/bindata/bootkube/manifests/configmap-kube-controller-manager-client-ca.yaml
+++ b/bindata/bootkube/manifests/configmap-kube-controller-manager-client-ca.yaml
@@ -6,5 +6,4 @@ metadata:
 data:
   ca-bundle.crt: |
     {{ .Assets | load "root-ca.crt" | indent 4 }}
-    {{ .Assets | load "kube-ca.crt" | indent 4 }}
 


s-urbaniak:
agreed :+1: having the kube-ca intermediate served by the cadvisor endpoint is the better approach. Let's hold this until the cadvisor patch lands.

deads2k:
> agreed  having the kube-ca intermediate served by the cadvisor endpoint is the better approach. Let's hold this until the cadvisor patch lands.

That concept seems to be moving in the wrong direction.  Upstream conformance for the service account token's ca.crt is that it is only guaranteed to be valid for the kube-apiserver when contacted via the service network.  We will not drift from upstream without an extremely compelling reason.  You'll need to communicate a different certificate.

mrogers950:
Generally this makes me question the purpose of the installer generating kube-ca, service-ca, etc. as intermediates of the root CA. The correct PKI behavior is that clients are provided a root and servers send their intermediates. However, we have the need for explicit trust domains based on a single CA, and a client only trusting an intermediate doesn't work out for all clients. If these intermediates were made as separate roots we would not have this issue, and we don't get any real benefit from them being intermediates..

sjenning:
I confirmed that after #112, Prometheus scraping of the kubelet and cadvisor metrics is working again.

The kubelet server cert is signed by the certificate given to the controller-manager via `--cluster-signing-cert-file` and `--cluster-signing-key-file` parameters.  Currently, that is the `kube-ca` intermediate and not the `root-ca`.

If the full chain is to be included in the cert, assuming the signing controller is using an intermediate CA, at would have to be added by the controller that generates the bundle in the CSR status once the CSR is approved.

abhinavdahiya:
> If these intermediates were made as separate roots we would not have this issue, and we don't get any real benefit from them being intermediates..

*i think the motivation for the intermediates was*

since the kube-ca.{crt,key} need to be in the cluster, if they are compromised (they have a greater chance of getting compromised), we need to rotate all the clients to trust the new kube-ca

with root-ca -> kube-ca setup , the root-ca.key never hits the cluster (hence smaller change of compromise) and we can simply rotate the kube-ca to new one.. and no change should be required on client's trust (although clients using certificate authentication will need to be rotated.)

mrogers950:
> since the kube-ca.{crt,key} need to be in the cluster, if they are compromised (they have a greater chance of getting compromised), we need to rotate all the clients to trust the new kube-ca

> with root-ca -> kube-ca setup , the root-ca.key never hits the cluster (hence smaller change of compromise) and we can simply rotate the kube-ca to new one.. and no change should be required on client's trust (although clients using certificate authentication will need to be rotated.)

This makes sense if all clients would trust just the root CA and servers always send their intermediate, but this does not work for when we want the trust to be separated explicitly between intermediates (i.e., a root CA loaded client could talk to a service serving certificate and also something from kube-ca). As we have found once we load only an intermediate this ends up breaking the (properly behaving) clients. 

Where does the root CA key end up? If you have to rotate kube-ca and expect it to stay an intermediate then you also need access to the root key to sign. If kube-ca were not an intermediate then it makes it easier for a rotation manager controller to rotate without need for access to an additional root CA key. Also, unless it is actually scrubbed from disk there is just as much of a chance of a full CA compromise, if not more so if it's left on a client.

I think we should have the installer self-sign these CAs instead. We would keep our trust separation and have an easier time with diverse clients and rotation/management. Not to mention, a server certificate has to have its intermediate included in its PEM file, which would make all secrets across the board much larger if we had to do that.

sttts:
/cc @smarterclayton @derekwaynecarr 

This seems to be a more fundamental discussion about our auth setup. How can we make progress here?

smarterclayton:
What did we do in 3.11?  Is this changing that behavior?  In 3.11 installer put the ca into every pod, I thought.

Regressions from security behavior around the core ca bundle probably are not acceptable without a massively good reason.  4.0 is not "break all compatibility with 3.x", it's "remove things we gave long notice for".

mrogers950:
@abhinavdahiya @smarterclayton Are you OK with my suggestion of changing the installer to create all the CAs as roots rather than intermediates? We have other teams that are now running into this problem with openssl/python clients. (/cc @shawn-hurley ).

PR #114, opened 2018-12-18
add service monitor resources for operator

deads2k:
/assign @mfojtik

### DIFF
--- a/manifests/0000_90_cluster-kube-controller-manager-operator_03_servicemonitor.yaml
+++ b/manifests/0000_90_cluster-kube-controller-manager-operator_03_servicemonitor.yaml
@@ -1,0 +1,27 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: cluster-kube-controller-manager-operator
+  namespace: openshift-cluster-kube-controller-manager-operator
+spec:
+  endpoints:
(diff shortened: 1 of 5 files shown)


manifests/0000_90_cluster-kube-controller-manager-operator_03_servicemonitor.yaml
```diff
@@ -0,0 +1,27 @@
...
+metadata:
+  name: cluster-kube-controller-manager-operator
+  namespace: openshift-cluster-kube-controller-manager-operator
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    interval: 30s
+    metricRelabelings:
+    - action: drop
+      regex: etcd_(debugging|disk|request|server).*
```
mfojtik: do we need this in controller manager?

mfojtik:
one nit, LGTM otherwise

manifests/0000_90_cluster-kube-controller-manager-operator_03_servicemonitor.yaml
deads2k: > do we need this in controller manager?

Because we're based on generic apiserver, I think we do.

PR #121, opened 2019-01-11
installer: QoS class guaranteed

sttts:
Peer of https://github.com/openshift/cluster-kube-apiserver-operator/pull/189.


sttts:
@aveshagarwal you might want this for the scheduler too.

PR #132, opened 2019-01-23
start rotating csr signing trust

deads2k:
Start rotating the CSR signing key.  This will back feed into the kube-apiserver for its client-ca.

/assign @mfojtik


deads2k:
/hold

we want https://github.com/openshift/cluster-kube-apiserver-operator/pull/218 first

sjenning:
There is no way to signal to the kubelet that they need to rotate their certs other than by using the expiration time on the cert.  Or by remove the client and server certs from the node, provisioning a new bootstrap cert for the kubelet to use, and restarting the kubelet.

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -0,0 +1,76 @@
...
+			Namespace:     operatorclient.OperatorNamespace,
+			Name:          "csr-controller-ca",
+			Informer:      kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().ConfigMaps(),
+			Lister:        kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().ConfigMaps().Lister(),
+			Client:        kubeClient.CoreV1(),
+			EventRecorder: eventRecorder,
+		},
+		certrotation.TargetRotation{
+			Namespace:         operatorclient.TargetNamespace,
+			Name:              "csr-signer",
```
deads2k: maybe this is a chain issue and I just need a new kind of target that back feeds the ca bundle

sjenning:
>  We trust the old signers until they expire.

ah goodly

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -0,0 +1,76 @@
...
+			Validity:          1 * 8 * time.Hour,
+			RefreshPercentage: 0.5,
+			Informer:          kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets(),
+			Lister:            kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets().Lister(),
+			Client:            kubeClient.CoreV1(),
+			EventRecorder:     eventRecorder,
+		},
+		certrotation.CABundleRotation{
+			Namespace:     operatorclient.OperatorNamespace,
+			Name:          "csr-controller-signer-ca",
```
sttts: csr-signer-ca ?

bindata/bootkube/manifests/configmap-initial-kube-controller-manager-csr-ca.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: initial-csr-ca
```
sttts: initial-csr-signer-ca

sttts:
Overall looks good. Please make the naming consistent, csr-signer-ca everywhere makes more sense IMO.

pkg/operator/certrotationcontroller/certrotationcontroller.go
deads2k: > csr-signer-ca

That is a ca bundle that includes the certs from the `csr-signer` so that we can rotate the signers quickly to prove we can while not disrupting trust

deads2k:
updated for comments and found a bug..  Fixed in a separate commit.

mfojtik:
/hold

```
E0131 11:55:29.004699       1 authentication.go:62] Unable to authenticate the request due to an error: [x509: certificate signed by unknown authority, x509: certificate signed by unknown authority (possibly because of "x509: invalid signature: parent certificate cannot sign this kind of certificate" while trying to verify candidate authority certificate "system:openshift:csr-signer")]
```

This is probably broken?

deads2k:
fixed and bumped

derekwaynecarr:
not sure if this was the cause, but all kubelets on workers are lost after ~12 hrs unable to authenticate back to apiserver due to bad certs.

derekwaynecarr:
need to verify kubelet is configured properly to rotate

sjenning:
yeah, my understanding was the old CA would still be trusted but this is apparently not the case :-(

```
# openssl x509 -in kubelet-server-current.pem -text -noout
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            76:3a:19:aa:d5:4a:a5:f9:1c:1c:3e:c0:98:b8:bc:fa:a9:17:81:00
    Signature Algorithm: sha256WithRSAEncryption
        Issuer: CN=kube-csr-signer_@1549295167
        Validity
            Not Before: Feb  4 15:47:00 2019 GMT
            Not After : Mar  6 15:47:00 2019 GMT
        Subject: O=system:nodes, CN=system:node:ip-10-0-134-227.us-west-1.compute.internal
...

# openssl x509 -in kubelet-client-current.pem -text -noout
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            68:31:6f:f5:3c:24:24:50:88:f6:f3:94:6e:3c:ef:9f:da:67:6f:8d
    Signature Algorithm: sha256WithRSAEncryption
        Issuer: CN=kube-csr-signer_@1549295167
        Validity
            Not Before: Feb  4 15:46:00 2019 GMT
            Not After : Mar  6 15:46:00 2019 GMT
        Subject: O=system:nodes, CN=system:node:ip-10-0-134-227.us-west-1.compute.internal
```
both client and server certs for the kubelet are for 30 days.

The cert manager in kube, which the kubelet uses, will rotate certs when they are 80%-/+10% used up.
For a 30 day cert that is after 21-27 days.  The kubelet will not rotate a cert that is still valid.

There is a `--experimental-cluster-signing-duration` that we likely need to set to something like 4 hours if we are going to rotate every 8 hours.

PR #135, opened 2019-01-28
Add OpenStack to ObserveCloudProviderNames

tomassedovic:
Running the operator on OpenStack currently results in "No recognized
cloud provider platform found in cloud config". This adds `openstack`
to the switch statement alongside libvirt and aws.


flaper87:
@eparis @crawford @wking Does this need an exception? Or can it be considered a bug fix?

It should unblock https://github.com/openshift/installer/issues/1092

flaper87:
@wking thanks! 

I can also confirm that this fixes OpenStack's deployment.

ingvagabund:
> We want this for openshift/installer#1107, which we want for openshift/installer#792, which is certainly a bug, so I'm comfortable calling this a bugfix.

/approve

PR #138, opened 2019-01-28
Disable persistent-local-volumes

gnufied:
Disable local-storage feature for now.

Fixes JIRA - STOR-183


deads2k:
@gnufied you probably want to set this at a higher level.  
@smarterclayton feature gate API I imagine

gnufied:
@deads2k https://github.com/openshift/cluster-kube-apiserver-operator/pull/228

deads2k:
> @deads2k [openshift/cluster-kube-apiserver-operator#228](https://github.com/openshift/cluster-kube-apiserver-operator/pull/228)

Also, you need to find the config overrides to disable this behavior on the bootstrap node

PR #141, opened 2019-01-30
bump(*): update library-go to add revision cm ownerrefs to installer/pruner pods

damemi:
bumps to add changes from ~~openshift/library-go#211~~ https://github.com/openshift/library-go/pull/215


sttts:
/retest

Prow is really broken.

PR #147, opened 2019-02-06
fix the csr-signer-ca bundle to keep history

deads2k:
We need to avoid removing the old certs we have from above.

/assign @sjenning


sjenning:
looks reasonable :+1:  testing now.

PR #150, opened 2019-02-07
reduce the sa/ca.crt to avoid trusting every CSR to impersonate the kube-apiserver

deads2k:
If the sa-token/ca.crt includes the trust bundles for the CSR endpoint, then the CSR endpoint becomes a high risk target for impersonating a kube-apiserver.  Since we are no longer rooting trust for the router wildcard, this is no longer necessary and in future release the upstream controller can distribute trust.

/assign @enj


enj:
YOLO

/lgtm

PR #152, opened 2019-02-11
switch to kube-ca as the starting ca.crt

deads2k:
This matches what we use from the "real" `sa/ca.rt`.

/assign @sttts @mrogers950


deads2k:
Going for this because it's closer.

deads2k:
Going for this because it's closer.

PR #153, opened 2019-02-11
generate our own sa token signer when missing

deads2k:
Instead of using the installer created SA token signer, create a new one.  This relies on https://github.com/openshift/origin/pull/21971 and https://github.com/openshift/cluster-kube-apiserver-operator/pull/238

/assign @sttts


deads2k:
/hold

this needs to wait until the endpoints list shows the bootstrap kube-apiserver (which won't respect it) gone

deads2k:
/hold cancel

@sttts new commit added

pkg/operator/certrotationcontroller/satokensigner_controller.go
```diff
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/sets"
```
sttts: nit: order

deads2k: got it

pkg/operator/certrotationcontroller/satokensigner_controller.go
```diff
@@ -91,6 +98,45 @@ func (c SATokenSignerController) sync() error {
...
+		for _, subset := range kubeEndpoints.Subsets {
+			for _, address := range subset.Addresses {
+				if !nodeIPs.Has(address.IP) {
+					unexpectedEndpoints.Insert(address.IP)
+				}
+			}
+		}
+		if len(unexpectedEndpoints) != 0 {
+			message := fmt.Sprintf("unexpected addresses:", unexpectedEndpoints.List())
+			c.eventRecorder.Warning("SATokenSignerControllerStuck", message)
```
sttts: is this really a warning? Should be normal during bootstrapping

sttts:
Looks good. Only https://github.com/openshift/cluster-kube-controller-manager-operator/pull/153#discussion_r257285104.

pkg/operator/certrotationcontroller/satokensigner_controller.go
deads2k: fixed

deads2k:
/hold

flaky as all get out.

deads2k:
/hold cancel

there is now a staging position until they pass the time allotted.  We can fix that later by having a better signal, but the code now tolerates waiting for that signal.

deads2k:
/hold 

holding for debug output

deads2k:
/retest

network

deads2k:
network

/retest

deads2k:
known flake on bootstrap user.  other failures were known networking flakes.

PR #154, opened 2019-02-11
rotate the kubeconfig for the controller manager

deads2k:
Instead of running with a kubeconfig file that is baked in during installation, use a kubeconfig based on values that are plumbed from the installer, use a kubeconfig maintained via operator and properly rotated.

/assign @sttts @mfojtik 

I suspect that we will end up having to expose a subset of service-ca in openshift-config-managed to wire through to here and the scheduler.

PR #155, opened 2019-02-12
Switch operator to use protobuf client for non-CR resources #250

mfojtik:
This builds on [openshift/library-go#233](https://github.com/openshift/library-go/pull/233) and switches the controller loops that deal with non-CR resources to client-go with protobuf enforced.

/cc @deads2k
/cc @smarterclayton


pkg/operator/starter.go
```diff
@@ -128,22 +128,28 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		ctx.EventRecorder,
 	)
 
-	certRotationController, err := certrotationcontroller.NewCertRotationController(kubeClient, operatorClient, kubeInformersForNamespaces, ctx.EventRecorder)
+	certRotationController, err := certrotationcontroller.NewCertRotationController(
+		kubeClient.CoreV1(),
```
deads2k: we want to use a cached one here, right?

mfojtik: yes, when library-go portion merge and I get it in bump

mfojtik:
seems to depend on https://github.com/openshift/cluster-kube-controller-manager-operator/pull/154

mfojtik:
/retest

installer flake

deads2k:
surprised this isn't conflicting, but sure

/lgtm

PR #156, opened 2019-02-12
 rotate the kubeconfig for the controller manager

deads2k:
if this works, it means that https://github.com/openshift/cluster-kube-controller-manager-operator/pull/154 isn't rebuilding the the image


mfojtik:
```
                   "message": "TargetConfigControllerFailing: \"v3.11.0/kube-controller-manager/leader-election-rolebinding.yaml\" (string): an empty namespace may not be set when a resource name is provided",
                        "reason": "TargetConfigControllerFailing",
```

PR #157, opened 2019-02-14
operator: remove openshift prefix from pod name

mfojtik:
We need https://github.com/openshift/installer/pull/1253 to support this transition.

In addition to changing the pod names, this also add revision suffix into pod container name, so it is clear in artifacts which revision the collected logs belonged to.

/hold

### DIFF
--- a/bindata/v3.11.0/kube-controller-manager/pod.yaml
+++ b/bindata/v3.11.0/kube-controller-manager/pod.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Pod
 metadata:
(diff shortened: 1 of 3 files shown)


bindata/v3.11.0/kube-controller-manager/pod.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     revision: "REVISION"
 spec:
   containers:
-  - name: openshift-kube-controller-manager
+  - name: kube-controller-manager-REVISION
```
deads2k: forgot to adjust the name above (line 4)?

PR #164, opened 2019-02-19
fix static pod start

deads2k:
https://github.com/openshift/library-go/pull/250

@sttts


deads2k:
bootstrap user, proxy, dc flakes. all known

PR #165, opened 2019-02-19
 add resource-graph for operator

deads2k:
Similar to KAS-o resource graph.  This shows what the KCM-o is doing.

resources 2


deads2k:
known flake.  adding a new command, merging

PR #166, opened 2019-02-19
observe cloud provider from config api

sanchezl:
Get cloud provider from `infrastructures.config.openshift.io/cluster` instead of cluster `cm/cluster-config-v1.kube-system`.


pkg/operator/starter.go
```diff
@@ -79,10 +80,12 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 	if err != nil {
 		return err
 	}
+	configInformers := configinformers.NewSharedInformerFactory(configClient, 10*time.Minute)
```
deads2k: move to where other clients and informers are created.

don't forget to start it.

sanchezl: Done.

deads2k:
known flake

PR #169, opened 2019-02-20
Fix file content detection race

deads2k:
proof of https://github.com/openshift/library-go/pull/267

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1678929

/assign @mfojtik 

@bparees it's a theory.

/hold

need to fix bump before merge


bparees:
it's a plausible theory + fix to me.
/lgtm
retag yourself post bump

PR #173, opened 2019-02-21
get network settings from config.openshift.io

sanchezl:
* Pickup new versions of `GetClusterCIDRs` & `GetServiceCIDR` from `github.com/openshift/library-go/pkg/operator/configobserver/network`.
* Remove `ConfigMapLister`


deads2k:
green and it is lgtm'd  bot is confused.

PR #174, opened 2019-02-22
bump(*): library-go get role binding fix and installer pod fix

mfojtik:


deads2k:
lgtm

ha proxy flake.

PR #175, opened 2019-02-22
feat(targetconfigcontroller): include service-ca bundle in SA cabundle

ericavonb:
Keep 3.11 compatibility by including the service CA from the service-serving-cert-signer controller in the service-account token secret.

For after: https://github.com/openshift/service-ca-operator/pull/20

Issue: [AUTH-131](https://jira.coreos.com/browse/AUTH-131)

cc @openshift/sig-auth @deads2k


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -237,6 +237,8 @@ func manageServiceAccountCABundle(lister corev1listers.ConfigMapLister, client c
 		// include the ca bundle needed to recognize default
 		// certificates generated by cluster-ingress-operator
 		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.GlobalMachineSpecifiedConfigNamespace, Name: "router-ca"},
+		// include ca bundle from service-ca
+		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.GlobalMachineSpecifiedConfigNamespace, Name: "service-ca"},
```
deads2k: The `ca.crt` should not contain the `service-ca` because the trust domains are different.  The service-ca.crt is a separate file because it is used for a different purpose and signs indiscriminately.  

This should match 3.x and place the service-ca into `service-ca.crt`.

ericavonb: Ok, that makes sense. Where is the right spot to wire that in?

ericavonb:
/close
in favor of #178

PR #176, opened 2019-02-22
switch to new discrete certs for sa/ca.crt

deads2k:
Depends on https://github.com/openshift/cluster-kube-apiserver-operator/pull/290 and https://github.com/openshift/installer/pull/1298 .

This brings in the trust published by the kube-apiserver


deads2k:
know cadvisor flake

PR #178, opened 2019-02-22
feat(targetconfigcontroller): include service-ca bundle in server account CA bundle

ericavonb:
### Bug  Report
🐛  Bug 1668825: serviceaccounttoken/service-ca.crt is missing

**Cause:** Missing feature to sync service-ca CA bundle to kube-controller-manager

**Consequence:** In 3.x, the service CA bundle for validating cluster-created
serving certificates was automatically mounted into pods via the ServiceAccount
token secret volume. Workloads that rely upon the service-ca.crt being present
would break in 4.0.

**Fix:** Sync the service CA bundle from the serving-cert-signer to the
kube-controller-manager where it can be injected into serviceaccount token
secrets at admission. The CA bundle will be provided by the service-ca-operator
in a configmap in the "openshift-config-managed" namespace. From there, the
cluster-kube-controller-manager-operator is responsible syncing it to the
kube-controller-manager.

**Result:** 3.x workloads can run on a 4.0 cluster and continue to utilize the
CA bundle found in the serviceaccounttoken/service-ca.crt file for validating
service certificates.

---

### Changes

🔗 AUTH-131: Deprecate SSCS ca in token secrets

1. **Add configmap syncer** to the config-observer controller
   Syncs the service CA bundle provided by the service-ca-operator in the
   "service-ca" configmap in the "openshift-config-managed" namespace to the
   "openshift-kube-controller-manager" namespace where it can be consumed by the
   kube-controller-manager.

2. **Mount the service CA bundle** data from the synced configmap into the
   kube-controller-manager pods at a well-known file location.

3. **Set "serviceServingCert.certFile"** field on the kube-controller-manager
   config to the file where the service-ca bundle is located if present.

### Testing
TODO @ericavonb

### DIFF
--- a/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
+++ b/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
@@ -13,6 +13,8 @@ import (
 	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/configobservation"
 	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/configobservation/cloudprovider"
 	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/configobservation/network"
+	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/configobservation/serviceca"
+	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/operatorclient"
 )
 
 type ConfigObserver struct {
@@ -23,6 +25,7 @@ func NewConfigObserver(
 	operatorClient v1helpers.OperatorClient,
 	operatorConfigInformers operatorv1informers.SharedInformerFactory,
 	configinformers configinformers.SharedInformerFactory,
+	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
 	resourceSyncer resourcesynccontroller.ResourceSyncer,
 	eventRecorder events.Recorder,
 ) *ConfigObserver {
@@ -34,17 +37,20 @@ func NewConfigObserver(
 				InfrastructureLister: configinformers.Config().V1().Infrastructures().Lister(),
 				NetworkLister:        configinformers.Config().V1().Networks().Lister(),
 				ResourceSync:         resourceSyncer,
+				ConfigMapLister:      kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps().Lister(),
 				PreRunCachesSynced: []cache.InformerSynced{
 					configinformers.Config().V1().Infrastructures().Informer().HasSynced,
 					configinformers.Config().V1().Networks().Informer().HasSynced,
+					kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps().Informer().HasSynced,
 				},
 			},
 			cloudprovider.ObserveCloudProviderNames,
 			network.ObserveClusterCIDRs,
 			network.ObserveServiceClusterIPRanges,
+			serviceca.ObserveServiceCA,
 		),
 	}
-
+	kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps().Informer().AddEventHandler(c.EventHandler())
 	operatorConfigInformers.Operator().V1().KubeControllerManagers().Informer().AddEventHandler(c.EventHandler())
 	configinformers.Config().V1().Infrastructures().Informer().AddEventHandler(c.EventHandler())
 	configinformers.Config().V1().Networks().Informer().AddEventHandler(c.EventHandler())
--- a/pkg/operator/configobservation/serviceca/observeserviceca.go
+++ b/pkg/operator/configobservation/serviceca/observeserviceca.go
@@ -1,0 +1,62 @@
+package serviceca
+
+import (
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/configobservation"
+	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/operatorclient"
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+const (
+	serviceCAConfigMapName = "service-ca"
+	serviceCABundleKey     = "ca-bundle.crt"
+	serviceCAFilePath      = "/etc/kubernetes/static-pod-resources/configmaps/service-ca/ca-bundle.crt"
+)
+
+// ObserveServiceCA fills in serviceServingCert.CertFile with the path for a configMap containing the service-ca.crt
+func ObserveServiceCA(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(configobservation.Listers)
+	errs := []error{}
+	prevObservedConfig := map[string]interface{}{}
+
+	topLevelServiceCAFilePath := []string{"serviceServingCert", "certFile"}
+	currentServiceCAFilePath, _, err := unstructured.NestedString(existingConfig, topLevelServiceCAFilePath...)
+	if err != nil {
+		errs = append(errs, err)
+	}
+	if len(currentServiceCAFilePath) > 0 {
+		if err := unstructured.SetNestedField(prevObservedConfig, currentServiceCAFilePath, topLevelServiceCAFilePath...); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	observedConfig := map[string]interface{}{}
+
+	// add service-ca configmap mount path to ServiceServingCert if configmap exists
+	ca, err := listers.ConfigMapLister.ConfigMaps(operatorclient.TargetNamespace).Get(serviceCAConfigMapName)
+	if errors.IsNotFound(err) {
+		// do nothing because we aren't going to add a path to a missing configmap
+		return observedConfig, errs
+	}
+	if err != nil {
+		// we had an error, return what we had before and exit. this really shouldn't happen
+		return prevObservedConfig, append(errs, err)
+	}
+	if len(ca.Data[serviceCABundleKey]) == 0 {
+		// do nothing because aren't going to add a path to a configmap with no file
+		return observedConfig, errs
+	}
+	// this means we have this configmap and it has values, so wire up the directory
+	if err := unstructured.SetNestedField(observedConfig, serviceCAFilePath, topLevelServiceCAFilePath...); err != nil {
+		recorder.Warningf("ObserveServiceCAConfigMap", "Failed setting serviceCAFile: %v", err)
+		errs = append(errs, err)
+	}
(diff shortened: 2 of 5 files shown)


pkg/operator/configobservation/serviceca/observeserviceca.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	if len(currentServiceCAFilePath) > 0 {
+		if err := unstructured.SetNestedField(prevObservedConfig, currentServiceCAFilePath, topLevelServiceCAFilePath...); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	// Sync the configMap in openshift-config-managed to the well-known resting place that corresponds to the service-ca path.
+	// If neither are set, this updates the destination with an empty source, which deletes the destination resource.
+	err = listers.ResourceSyncer().SyncConfigMap(
```
deads2k: you could add this statically to the resourcesynccontroller file.  there is no variation in the name.

pkg/operator/configobservation/serviceca/observeserviceca.go
```diff
@@ -0,0 +1,76 @@
...
+	serviceCAConfigMapName = "service-ca"
+)
+
+// ObserveServiceCA fills in serviceServingCert.CertFile with the path for a configMap containing the service-ca.crt
+func ObserveServiceCA(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(configobservation.Listers)
+	errs := []error{}
+	prevObservedConfig := map[string]interface{}{}
+
+	topLevelServiceCAFilePath := []string{"service-ca", "service-ca.crt"}
```
deads2k: `serviceServingCert.certFile` isn't it?

pkg/operator/configobservation/serviceca/observeserviceca.go
```diff
@@ -0,0 +1,76 @@
...
+	if len(ca.Data) <= 0 {
+		// do nothing because aren't going to add a path to a configmap with no files
+		return observedConfig, errs
+	}
+	// this means we have this configmap and it has values, so wire up the directory
+	if err := unstructured.SetNestedField(observedConfig, serviceCAFilePath, topLevelServiceCAFilePath...); err != nil {
+		recorder.Eventf("ObserveServiceCAConfigMap", "Failed setting serviceCAFile: %v", err)
+		errs = append(errs, err)
+	}
+
```
deads2k: diff the new and existing values to emit an event when you get a change

pkg/operator/configobservation/serviceca/observeserviceca.go
```diff
@@ -0,0 +1,76 @@
...
+		// we had an error, return what we had before and exit. this really shouldn't happen
+		return prevObservedConfig, append(errs, err)
+	}
+	if len(ca.Data) <= 0 {
+		// do nothing because aren't going to add a path to a configmap with no files
+		return observedConfig, errs
+	}
+	// this means we have this configmap and it has values, so wire up the directory
+	if err := unstructured.SetNestedField(observedConfig, serviceCAFilePath, topLevelServiceCAFilePath...); err != nil {
+		recorder.Eventf("ObserveServiceCAConfigMap", "Failed setting serviceCAFile: %v", err)
```
deads2k: this seems like a warning.

pkg/operator/starter.go
```diff
@@ -176,6 +177,7 @@ var deploymentConfigMaps = []revision.RevisionResource{
 	{Name: "config"},
 	{Name: "controller-manager-kubeconfig"},
 	{Name: "serviceaccount-ca"},
+	{Name: "service-ca", Optional: true},
```
deads2k: if you're checking length and statically sync, then you don't need to be optional.

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -34,6 +37,7 @@ func NewConfigObserver(
 				InfrastructureLister: configinformers.Config().V1().Infrastructures().Lister(),
 				NetworkLister:        configinformers.Config().V1().Networks().Lister(),
 				ResourceSync:         resourceSyncer,
+				ConfigMapLister:      configMapLister,
```
deads2k: looks like you forgot to wait for a sync and you forgot to wire up to the informer to add an event handler.

pkg/operator/configobservation/serviceca/observeserviceca.go
ericavonb: Does it have to match exactly? I figured the file path would be independent of the config field.

pkg/operator/configobservation/serviceca/observeserviceca.go
ericavonb: f83f0b49
(will rebase and squash commits before merge)

pkg/operator/configobservation/serviceca/observeserviceca.go
ericavonb: f204e00f

pkg/operator/configobservation/serviceca/observeserviceca.go
ericavonb: 3704a612 plz advise if there's a better event message or structure to follow

pkg/operator/starter.go
ericavonb: bc12f357

pkg/operator/configobservation/serviceca/observeserviceca.go
ericavonb: ignore me: 53e7e279

pkg/operator/configobservation/serviceca/observeserviceca.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	if len(currentServiceCAFilePath) > 0 {
+		if err := unstructured.SetNestedField(prevObservedConfig, currentServiceCAFilePath, topLevelServiceCAFilePath...); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	// Sync the configMap in openshift-config-managed to the well-known resting place that corresponds to the service-ca path.
+	// If neither are set, this updates the destination with an empty source, which deletes the destination resource.
+	if err != nil {
```
enj: I think this got left in from an earlier commit.

ericavonb: 🤦‍♀️ a257842a

pkg/operator/configobservation/serviceca/observeserviceca.go
```diff
@@ -0,0 +1,61 @@
...
+	ca, err := listers.ConfigMapLister.ConfigMaps(operatorclient.TargetNamespace).Get(serviceCAConfigMapName)
+	if errors.IsNotFound(err) {
+		// do nothing because we aren't going to add a path to a missing configmap
+		return observedConfig, errs
+	}
+	if err != nil {
+		// we had an error, return what we had before and exit. this really shouldn't happen
+		return prevObservedConfig, append(errs, err)
+	}
+	if len(ca.Data) <= 0 {
```
enj: I think we want to make sure the expected key has data.

ericavonb: 4957ab4f I'm unsure of the behavior on the kube-controller-manager side if the key is set to the empty string.

pkg/operator/starter.go
```diff
@@ -176,6 +177,7 @@ var deploymentConfigMaps = []revision.RevisionResource{
 	{Name: "config"},
 	{Name: "controller-manager-kubeconfig"},
 	{Name: "serviceaccount-ca"},
+	{Name: "service-ca"},
```
mrogers950: This might need to be set with Optional: true since it's not an installer created resource

ericavonb: I thought so too: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/178#discussion_r259874110

pkg/operator/configobservation/serviceca/observeserviceca.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/configobservation"
+	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/operatorclient"
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+const (
+	serviceCAConfigMapName = "service-ca"
+	serviceCABundleKey     = "cabundle.crt"
```
enj: OAuth server and SA stuff uses `ca.crt`.  I also see lots of `ca-bundle.crt`.  We should probably pick one of those?

deads2k: Update the other side to produce `ca-bundle.crt` and update here.  `must-gather` also understands `ca-bundle.crt`

pkg/operator/configobservation/serviceca/observeserviceca.go
```diff
@@ -0,0 +1,62 @@
...
+	ca, err := listers.ConfigMapLister.ConfigMaps(operatorclient.TargetNamespace).Get(serviceCAConfigMapName)
+	if errors.IsNotFound(err) {
+		// do nothing because we aren't going to add a path to a missing configmap
+		return observedConfig, errs
+	}
+	if err != nil {
+		// we had an error, return what we had before and exit. this really shouldn't happen
+		return prevObservedConfig, append(errs, err)
+	}
+	if _, ok := ca.Data[serviceCABundleKey]; !ok {
```
deads2k: don't add it if the file is empty.

pkg/operator/configobservation/serviceca/observeserviceca.go
deads2k: > OAuth server and SA stuff uses `ca.crt`. I also see lots of `ca-bundle.crt`. We should probably pick one of those?

ca-bundle.crt

ericavonb: https://github.com/openshift/service-ca-operator/pull/27

ericavonb: 0222f146 - should start passing after https://github.com/openshift/service-ca-operator/pull/27

pkg/operator/configobservation/serviceca/observeserviceca.go
ericavonb: b85936ad

ericavonb:
@deads2k addressed your comments. please take another look

pkg/operator/configobservation/serviceca/observeserviceca.go
```diff
@@ -0,0 +1,62 @@
...
+	ca, err := listers.ConfigMapLister.ConfigMaps(operatorclient.TargetNamespace).Get(serviceCAConfigMapName)
+	if errors.IsNotFound(err) {
+		// do nothing because we aren't going to add a path to a missing configmap
+		return observedConfig, errs
+	}
+	if err != nil {
+		// we had an error, return what we had before and exit. this really shouldn't happen
+		return prevObservedConfig, append(errs, err)
+	}
+	if len(ca.Data[serviceCABundleKey]) < 1 {
```
enj: `len(...) == 0`

deads2k: > len(...) == 0

agree

deads2k:
update length comparison to the project standard and this lgtm

/approve

pkg/operator/configobservation/serviceca/observeserviceca.go
ericavonb: 92fd9e39

PR #177, opened 2019-02-22
 adopt the CSR from the installer

deads2k:
Pick up the CSR from the installer and start rotating based on it.

PR #180, opened 2019-02-25
bump(*): library-go pickup for rolebinding fix

mfojtik:


deads2k:
I would like to get https://github.com/openshift/cluster-kube-controller-manager-operator/pull/177 first. It should pass after https://github.com/openshift/cluster-kube-apiserver-operator/pull/301 merges

PR #197, opened 2019-03-22
pod.yaml: add init container waiting for free port

soltysh:
/assign @sttts

### DIFF
--- a/bindata/v3.11.0/kube-controller-manager/pod.yaml
+++ b/bindata/v3.11.0/kube-controller-manager/pod.yaml
@@ -8,6 +8,18 @@ metadata:
     kube-controller-manager: "true"
     revision: "REVISION"
 spec:
+  initContainers:
+  - name: wait-for-host-port
+    image: ${IMAGE}
+    imagePullPolicy: IfNotPresent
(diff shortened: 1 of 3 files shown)


bindata/v3.11.0/kube-controller-manager/pod.yaml
```diff
@@ -8,6 +8,18 @@ metadata:
     kube-controller-manager: "true"
     revision: "REVISION"
 spec:
+  initContainers:
+  - name: wait-for-host-port
+    image: ${IMAGE}
+    imagePullPolicy: IfNotPresent
+    command: ['/usr/bin/timeout', '100', "/bin/bash", "-c"] # 60 sec for graceful termination, 35 for minimum-termination-duration
```
soltysh: @sttts I've copied 100 here, if you know any reason this should be shorter lemme know

sttts: It can. We don't have graceful termination for kube-controller-manager. 10 or at most 30 secs would be a good number.

mfojtik:
cleared via Slack... in this PR we still see controller manager trying to bind on the insecure port, @soltysh have PR that fixes wiring the flags, we can let this go and the other PR should fix passing the flags.

/hold cancel

PR #199, opened 2019-03-25
specify kubeconfig for delegated authn and authz

deads2k:
Pretty sure this is causing the rebase KCM to fail.


bindata/v3.11.0/kube-controller-manager/pod.yaml
```diff
@@ -17,6 +17,8 @@ spec:
     args:
     - --openshift-config=/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml
     - --kubeconfig=/etc/kubernetes/static-pod-resources/configmaps/controller-manager-kubeconfig/kubeconfig
+    - --authentication-kubeconfig=/etc/kubernetes/static-pod-resources/configmaps/controller-manager-kubeconfig/kubeconfig
```
sttts: do we really need config and flags?

sttts:
Rebased in https://github.com/openshift/cluster-kube-controller-manager-operator/pull/213.

PR #200, opened 2019-03-25
disable kcm secure serving

deads2k:
We are not generating serving certs and we do not tolerate not having these certs, so this disables it until we sort out what to do.  This is why the KCM is dying in the rebase.


deads2k:
approved via slack.  Needed for rebase

sttts:
@soltysh @deads2k is https://github.com/kubernetes/kubernetes/pull/69884 related?

soltysh:
It looks like, we should set `--cert-dir` or `--tls-cert-file` and `--tls-private-key-file` to have those used.

PR #203, opened 2019-03-27
Add cert-rotation configmap for setting base rotation interval

soltysh:
/assign @deads2k


pkg/operator/starter.go
```diff
@@ -139,12 +140,31 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		ctx.EventRecorder,
 	)
 
+	certRotationBase := time.Duration(0)
+	err = wait.PollImmediate(time.Second, 1*time.Minute, func() (bool, error) {
```
deads2k: has this leaked other places?  smells like a library function if we're going to do it this way.

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -22,17 +27,25 @@ func NewCertRotationController(
...
+
 	certRotator, err := certrotation.NewCertRotationController(
 		"CSRSigningCert",
 		certrotation.SigningRotation{
 			Namespace: operatorclient.OperatorNamespace,
 			// this is not a typo, this is the signer of the signer
 			Name:          "csr-signer-signer",
-			Validity:      8 * time.Hour, // to be 10 days
-			Refresh:       4 * time.Hour, // to be 4 days
+			Validity:      30 * rotationDay,
```
deads2k: I think this one needs to be 60 days and 30 days

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -49,8 +62,8 @@ func NewCertRotationController(
 		certrotation.TargetRotation{
 			Namespace: operatorclient.OperatorNamespace,
 			Name:      "csr-signer",
-			Validity:  1 * 4 * time.Hour, // to be 5 days
-			Refresh:   2 * time.Hour,     // to be 1 day
+			Validity:  30 * rotationDay,
+			Refresh:   15 * rotationDay,
```
deads2k: any reason not to rotate this every day?

deads2k: > any reason not to rotate this every day?

oh, you're chasing the restart.  Someone look to see if we can just make it reload serving trust dynamically?

pkg/operator/starter.go
```diff
@@ -139,12 +140,31 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
+		certRotationConfig, err := kubeClient.CoreV1().ConfigMaps(operatorclient.GlobalUserSpecifiedConfigNamespace).Get("unsupported-cert-rotation-config", metav1.GetOptions{})
+		if err != nil {
+			if errors.IsNotFound(err) {
+				return true, nil
```
mfojtik: is this just retrying for connection errors?

pkg/operator/starter.go
soltysh: Yes, we have identical code in kas-o's starter.go

pkg/operator/starter.go
soltysh: That is correct.

pkg/operator/certrotationcontroller/certrotationcontroller.go
soltysh: Discussed on slack, I'll be looking into that dynamic reload of kcm certs in kas.

pkg/operator/starter.go
```diff
@@ -139,12 +140,31 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
+			}
+			return false, err
+		}
+		if value, ok := certRotationConfig.Data["base"]; ok {
+			certRotationBase, _ = time.ParseDuration(value)
```
deads2k: make this fail, so we know

PR #204, opened 2019-03-27
cloudprovider: add vsphere provider

mfojtik:


pkg/operator/configobservation/cloudprovider/observe_cloudprovider.go
```diff
@@ -44,6 +44,8 @@ func ObserveCloudProviderNames(genericListers configobserver.Listers, recorder e
 		return previouslyObservedConfig, errs
 	case configv1.AWSPlatform:
 		cloudProvider = "aws"
+	case configv1.VSpherePlatform:
```
ingvagabund: Is there an issue linking all `vsphere` related PRs into a single place? Just for reference. Otherwise, looks good.

ingvagabund: `vsphere` intra platform got introduced by https://github.com/openshift/api/pull/182. Though, it's quite generic PR.

PR #206, opened 2019-04-03
Sync cloudprovider configmap

gnufied:
This PR still depends on https://github.com/openshift/installer/pull/1516 for actual `Infrastructure` object to be populated. 

cc @deads2k

### DIFF
--- a/pkg/operator/configobservation/cloudprovider/observe_cloudprovider.go
+++ b/pkg/operator/configobservation/cloudprovider/observe_cloudprovider.go
@@ -9,15 +9,23 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/library-go/pkg/operator/configobserver"
 	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
 
 	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/configobservation"
 )
 
-// ObserveCloudProviderNames observes the loud provider from the global cluster infrastructure resource.
+const (
+	targetNamespaceName       = "openshift-kube-controller-manager"
+	cloudProviderConfFilePath = "/etc/kubernetes/static-pod-resources/configmaps/cloud-config/config"
+	configNamespace           = "openshift-config"
+)
+
+// ObserveCloudProviderNames observes the cloud provider from the global cluster infrastructure resource.
 func ObserveCloudProviderNames(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
 	listers := genericListers.(configobservation.Listers)
 	var errs []error
 	cloudProvidersPath := []string{"extendedArguments", "cloud-provider"}
+	cloudProviderConfPath := []string{"extendedArguments", "cloud-config"}
 
 	previouslyObservedConfig := map[string]interface{}{}
 	if currentCloudProvider, _, _ := unstructured.NestedStringSlice(existingConfig, cloudProvidersPath...); len(currentCloudProvider) > 0 {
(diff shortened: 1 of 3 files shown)


gnufied:
cc @staebler @openshift/sig-storage

pkg/operator/configobservation/cloudprovider/observe_cloudprovider.go
```diff
@@ -9,15 +9,23 @@ import (
...
+	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
 
 	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/configobservation"
 )
 
-// ObserveCloudProviderNames observes the loud provider from the global cluster infrastructure resource.
+const (
+	targetNamespaceName       = "openshift-kube-controller-manager"
+	cloudProviderConfFilePath = "/etc/kubernetes/static-pod-resources/configmaps/cloud-config/config"
+	configNamespace           = "openshift-config"
```
deads2k: @derekwaynecarr I think we want this to be `openshift-config-managed` to be consistent for all clouds and allow future filtering, right?

pkg/operator/configobservation/cloudprovider/observe_cloudprovider.go
```diff
@@ -26,22 +34,62 @@ func ObserveCloudProviderNames(genericListers configobserver.Listers, recorder e
...
+	cloudProvider := getPlatformName(infrastructure.Status.Platform, recorder)
+	if len(cloudProvider) > 0 {
+		if err := unstructured.SetNestedStringSlice(observedConfig, []string{cloudProvider}, cloudProvidersPath...); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	sourceCloudConfigMap := infrastructure.Spec.CloudConfig.Name
+	sourceCloudConfigNamespace := configNamespace
+	if len(sourceCloudConfigMap) == 0 {
```
deads2k: In this case, I think we need to sync an empty location to clear the destination.  Can you have a look at the godoc on syncconfigmap?

pkg/operator/configobservation/cloudprovider/observe_cloudprovider.go
```diff
@@ -26,22 +34,62 @@ func ObserveCloudProviderNames(genericListers configobserver.Listers, recorder e
...
+	)
+	if err != nil {
+		errs = append(errs, err)
+		return observedConfig, errs
+	}
+	if err := unstructured.SetNestedStringSlice(observedConfig, []string{cloudProviderConfFilePath}, cloudProviderConfPath...); err != nil {
+		recorder.Warningf("ObserverCloudProviderNames", "Failed setting cloud-config : %v", err)
+		errs = append(errs, err)
+	}
+	return observedConfig, errs
```
deads2k: just before returning, can you look for a diff between the prevconfig and the newconfig and emit an event.  @mfojtik probably has an example.

pkg/operator/configobservation/cloudprovider/observe_cloudprovider.go
gnufied: Okay, if this has to change https://github.com/openshift/installer/pull/1516/files#diff-8db489d59cedb0a4f04c7546779f58a2R68 should be updated to reflect new location right?

pkg/operator/configobservation/cloudprovider/observe_cloudprovider.go
gnufied: fixed

pkg/operator/configobservation/cloudprovider/observe_cloudprovider.go
gnufied: ^ does this work or did you mean deep comparsion of actual cloudprovider configuration?

pkg/operator/configobservation/cloudprovider/observe_cloudprovider.go
staebler: The documentation of Infrastructure should change, too, then.
https://github.com/openshift/api/blob/d2f01e7b77a6fc78b328db20285423838419fef7/config/v1/types_infrastructure.go#L28

pkg/operator/configobservation/cloudprovider/observe_cloudprovider.go
```diff
@@ -3,45 +3,114 @@ package cloudprovider
...
+// ObserveCloudProviderNames observes the cloud provider from the global cluster infrastructure resource.
 func ObserveCloudProviderNames(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
 	listers := genericListers.(configobservation.Listers)
 	var errs []error
 	cloudProvidersPath := []string{"extendedArguments", "cloud-provider"}
-
+	cloudProviderConfPath := []string{"extendedArguments", "cloud-config"}
 	previouslyObservedConfig := map[string]interface{}{}
+
+	existinCloudConfig, _, err := unstructured.NestedStringSlice(existingConfig, cloudProviderConfPath...)
```
staebler: s/existinCloudConfig/existingCloudConfig/

pkg/operator/configobservation/cloudprovider/observe_cloudprovider.go
```diff
@@ -3,45 +3,114 @@ package cloudprovider
...
+func getPlatformName(platformType configv1.PlatformType, recorder events.Recorder) string {
 	cloudProvider := ""
-	switch infrastructure.Status.Platform {
+	switch platformType {
 	case "":
 		recorder.Warningf("ObserveCloudProvidersFailed", "Required status.platform field is not set in infrastructures.%s/cluster", configv1.GroupName)
-		return previouslyObservedConfig, errs
 	case configv1.AWSPlatform:
 		cloudProvider = "aws"
 	case configv1.VSpherePlatform:
```
awesomenix: Can you please add support for Azure platform as well

gnufied: I can't add support for Azure in this PR. Because it was decided that applying a opaque configmap as a cloudprovider config will only be supported for vsphere for the time being. For other cloudproviders, plan is to validate and support only options we know work.

gnufied:
I have verified and validated that vsphere cloudprovider configuration for controller-manager in 4.1 works after this PR. I was able to provision and attach/detach PVCs successfully in a working cluster. 

@derekwaynecarr @deads2k PTAL.

PR #207, opened 2019-04-03
switch controller manager to secure port

mfojtik:
Rebase landed, we should switch the port to secure again. This updates controller manager default config and set the insecure port to 0 and also update the init container to check the secure port.

Also sets the `cert-dir` to `/var/run/kubernetes`.


mfojtik:
adding lgtm back, fixed the liveness probe port

soltysh:
/hold 
we need https://github.com/openshift/origin/pull/22476 first and than we can land this.

sttts:
/assign @s-urbaniak 

This also needs a change on the monitoring side.

PR #211, opened 2019-04-04
Add support for Azure platform

awesomenix:
Add support for azure platform cloud provider parameter


gnufied:
Has anyone validated Azure cloudprovider configured with controller-manager? Do we need anything beyond this PR?

PR #212, opened 2019-04-05
bindata: use default node controller settings

sjenning:
@mfojtik @enxebre @derekwaynecarr 

We should not be changing ` node-monitor-grace-period` and `pod-eviction-timeout` from the defaults without cause and there is no cause provided.


enxebre:
My concern with that missing context on why this was originally set to 5. Hopefully reviewers can shed some light

enxebre:
To give more context: The node-monitor-grace-period ("Amount of time which we allow starting Node to be unresponsive before marking it unhealthy.") https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/nodelifecycle/node_lifecycle_controller.go#L887 is set to 5 min  so node status won't report unhealthy until 5 min after e.g a machine is shut down manually

PR #215, opened 2019-04-09
 add observer for feature gates

deads2k:
rebump and add the library-go feature gate function


deads2k:
e2e is almost done.  If it's green, I'll push one on top of this and force merge after green verify.  If it's not, I'll push and wait.

eparis:
https://github.com/openshift/api/pull/278

PR #218, opened 2019-04-11
manifests: extended wait for port time to 105s

mfojtik:


mfojtik:
/hold

we might not need to do this if we close the ports properly https://github.com/openshift/origin/pull/22543

ravisantoshgudimetla:
I am testing this PR to see if resourcequota featuregate PR is having any effect.

PR #230, opened 2019-04-18
Use the CVO to create the operator config

damemi:
Same changes as https://github.com/openshift/cluster-kube-apiserver-operator/pull/418


manifests/0000_12_kube-controller-manager-operator_01_operator.cr.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: operator.openshift.io/v1
+kind: KubeControllerManager
+metadata:
+  name: cluster
+  annotations:
+    release.openshift.io/create-only: "true"
```
ravisantoshgudimetla: Aren't you missing replicas/logging here? I know log-level by default is 2 which should be ok but what about replicas?
@damemi

PR #233, opened 2019-04-22
Add logging, replicas back to operator manifest

damemi:
As https://github.com/openshift/cluster-kube-controller-manager-operator/pull/230/files#r277413054 noticed, we dropped the logging and replicas when moving the manifest from bindata. This adds it back in


damemi:
cc @deads2k @ravisantoshgudimetla do we know we need this?

deads2k:
we don't use these fields

/close

PR #238, opened 2019-04-26
bump(*): library-go to include revision controller deletion recovery

damemi:
Includes changes from openshift/library-go#368 to recover from operator resource deletion


damemi:
already bumped

PR #240, opened 2019-04-29
change key for cluster-operator version

deads2k:
This allows us to separate changes to the operator level from changes in the overall level.

/assign @mfojtik


deads2k:
tagging in case it goes green in the next 20 minutes

PR #246, opened 2019-05-02
Bug 1705649 : Cluster with halted master did not reschedule operators after 5m of being down

ravisantoshgudimetla:
As of now, because of infinite tolerations against all the possible taints, we are seeing that operators are not getting evicted from nodes that have NoExecute taint on them. This PR tightens the conditions around which can be pods can be scheduled/evicted. The downside is there is a very good chance that pods would be evicted from nodes that have certain conditions like disk-pressure, memory-pressure, taints added by other controllers(operators) etc. So, please make sure that this change is ok with your operator/operand before merging this PR.

/cc @sjenning @smarterclayton


ravisantoshgudimetla:
/hold

I want to remove the tolerations from static pods.

PR #254, opened 2019-05-17
Restart controller manager process when certs change on disk

mfojtik:
This is added a side-car container to the controller-manager process that use shared PID namespace to watch the certificates in the main container and perform graceful termination of the main container (controller-manager) process when the certificates change on disk.

/cc @tnozicka 
/cc @deads2k 
/cc @sttts


mfojtik:
We already redeploy when these certs change, closing.

PR #255, opened 2019-05-17
Bump operator dependencies to 1.14.0

mfojtik:


mfojtik:
/hold

we need to fix the crd gen first.

mfojtik:
adding approved label manually, for some reason it did now worked.

PR #262, opened 2019-06-17
bindata/kube-controller-manager: fix cluster-monitoring label

s-urbaniak:
Currently `openshift.io/cluster-monitoring` is applied as an annotation,
it should be a label.

Currently, in a freshly deployed cluster I don't even see the annotation applied though at all. The piece of code that does that (I believe) is here: https://github.com/openshift/cluster-kube-controller-manager-operator/blob/f3b13ef52ab480675cb6a665c1b4746a187db7d7/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L134

Let's see if labels get propagated successfully.

/assign @soltysh @mfojtik

### DIFF
--- a/bindata/bootkube/manifests/00_openshift-kube-controller-manager-ns.yaml
+++ b/bindata/bootkube/manifests/00_openshift-kube-controller-manager-ns.yaml
@@ -3,7 +3,7 @@ kind: Namespace
 metadata:
   annotations:
     openshift.io/node-selector: ""
-    openshift.io/cluster-monitoring: "true"
   name: openshift-kube-controller-manager
   labels:
     openshift.io/run-level: "0"
+    openshift.io/cluster-monitoring: "true"
--- a/bindata/v3.11.0/kube-controller-manager/ns.yaml
+++ b/bindata/v3.11.0/kube-controller-manager/ns.yaml
@@ -3,7 +3,7 @@ kind: Namespace
 metadata:
   annotations:
     openshift.io/node-selector: ""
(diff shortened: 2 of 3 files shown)


s-urbaniak:
I need to look into this, Prometheus doesn't seem to be happy :thinking: 
```
fail [github.com/openshift/origin/test/extended/prometheus/prometheus_builds.go:167]: Expected
    <map[string]error | len:1>: {
        "sum(ALERTS{alertstate=\"firing\"})": {
            s: "query sum(ALERTS{alertstate=\"firing\"}) for tests []prometheus.metricTest{prometheus.metricTest{labels:map[string]string(nil), greaterThanEqual:false, value:2, success:false}} had results {\"status\":\"success\",\"data\":{\"resultType\":\"vector\",\"result\":[{\"metric\":{},\"value\":[1560797697.887,\"3\"]}]}}",
        },
    }
to be empty
```

s-urbaniak:
found, why the e2e tests are failing:

image

With this PR we have target duplicates, and this bubbles up as alerts, causing the e2e failure. Can I assume that temporarily ignoring specifically the `PrometheusTargetScrapesDuplicate` alert for the time being of this transition is safe? /cc @brancz @paulfantom ?

brancz:
I didn’t even realize we have that. Yes we should temporarily disable.

brancz:
Good thing is the alert is telling us that we are doing the right thing :)

s-urbaniak:
@brancz definitely ;-) i was wondering why this fails actually.

PR #265, opened 2019-07-10
Use new crd-schema-gen Makefile pattern

damemi:
This follows the same approach as
* https://github.com/openshift/cluster-kube-apiserver-operator/pull/514
* https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/105


manifests/0000_25_kube-controller-manager-operator_01_config.crd.yaml
```diff
@@ -183,5 +559,4 @@ spec:
               description: version is the level this availability applies to
               type: string
           type: object
-      required:
-      - spec
```
sttts: did we change that?

damemi: uh oh, https://github.com/openshift/api/blob/master/operator/v1/types_kubecontrollermanager.go#L16

we use `// +required` everywhere, but that isn't a controller-tools marker (because before, everything was required. So we created `// +kubebuilder:validation:Required`). Using `kubebuilder:validation:Required` tag works, and adds the spec to required. I can probably add `// +required` upstream for parity

damemi: Hopefully no one tries to create an operator without a spec, because we've merged this already into kas and kcm. If they do, that is the fix

damemi: https://github.com/kubernetes-sigs/controller-tools/pull/247

PR #268, opened 2019-07-15
vendor: bump to get update cloudproviderconfig observer with GCP

abhinavdahiya:
update from https://github.com/openshift/library-go/pull/473


glide.yaml
```diff
@@ -15,7 +15,7 @@ import:
 - package: k8s.io/component-base
   version: release-1.14
 - package: github.com/openshift/api
-  version: master
+  version: cf189d3400156fba0afb233aa8fca57a7ff61cb2
```
abhinavdahiya: https://github.com/openshift/api/pull/370 seems to have broken the library-go code configobserver for featuregates.

i'm not sure of the code there, but fixing to version before https://github.com/openshift/api/pull/370 seems to be working.
need this gce cloudprovider bump to unblock GCP work.. @deads2k how should we proceed

deads2k: If library-go needs an update,then it should be updated.

deads2k: > [openshift/api#370](https://github.com/openshift/api/pull/370) seems to have broken the library-go code configobserver for featuregates.
> 
> i'm not sure of the code there, but fixing to version before [openshift/api#370](https://github.com/openshift/api/pull/370) seems to be working.
> need this gce cloudprovider bump to unblock GCP work.. @deads2k how should we proceed

@p0lyn0mial had https://github.com/openshift/library-go/pull/474 .  it's close enough that I merged it.

PR #269, opened 2019-07-18
Add observer for the cluster's infra ID and set it as the cluster-name argument

csrwng:
Adds cluster-name argument to extendedArguments with a value of the cluster's infra ID.

### DIFF
--- a/pkg/operator/configobservation/clustername/observeinfraid.go
+++ b/pkg/operator/configobservation/clustername/observeinfraid.go
@@ -1,0 +1,45 @@
+package clustername
+
+import (
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/configobservation"
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+// ObserveInfraID fills in the cluster-name extended argument for the controller-manager with the cluster's infra ID
+func ObserveInfraID(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(configobservation.Listers)
+	errs := []error{}
+	clusterNamePath := []string{"extendedArguments", "cluster-name"}
+	previouslyObservedConfig := map[string]interface{}{}
+
+	if currentClusterName, _, _ := unstructured.NestedStringSlice(existingConfig, clusterNamePath...); len(currentClusterName) > 0 {
+		if err := unstructured.SetNestedStringSlice(previouslyObservedConfig, currentClusterName, clusterNamePath...); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	observedConfig := map[string]interface{}{}
+	infrastructure, err := listers.InfrastructureLister().Get("cluster")
+	if err != nil {
+		if errors.IsNotFound(err) {
+			recorder.Warningf("ObserveInfraID", "Required infrastructures.%s/cluster not found", configv1.GroupName)
+		}
+		return previouslyObservedConfig, errs
+	}
+	// The infrastructureName value in infrastructure status is always present and cannot be changed during the
+	// lifetime of the cluster.
+	infraID := infrastructure.Status.InfrastructureName
+	if len(infraID) == 0 {
+		recorder.Warningf("ObserveInfraID", "Value for infrastructureName in infrastructure.%s/cluster is blank", configv1.GroupName)
+		return previouslyObservedConfig, errs
+	}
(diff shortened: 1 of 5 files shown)


csrwng:
@deads2k added the observer however not sure how to block startup if it's not set, ptal

csrwng:
> `listing hosted zones: Throttling: Rate exceeded`

/test e2e-aws-operator

csrwng:
Added code to prevent startup of the controller manager if the cluster-name argument is not present.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -126,6 +136,40 @@ func (c TargetConfigController) sync() error {
 	return nil
 }
 
+func isRequiredConfigPresent(config []byte) error {
```
sttts: cluster-kube-apiserver-operator also has tests for this. Better copy them as well.

csrwng: done

csrwng:
@sttts bump. Are any more changes needed? This is needed for GCP support.

pkg/operator/configobservation/clustername/observeinfraid.go
```diff
@@ -0,0 +1,40 @@
...
+		if err := unstructured.SetNestedStringSlice(previouslyObservedConfig, currentClusterName, clusterNamePath...); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	observedConfig := map[string]interface{}{}
+	infrastructure, err := listers.InfrastructureLister().Get("cluster")
+	if errors.IsNotFound(err) {
+		recorder.Warningf("ObserveInfraID", "Required infrastructures.%s/cluster not found", configv1.GroupName)
+		return observedConfig, errs
```
deads2k: you sure you want to clear the value instead of using the previous one?

pkg/operator/configobservation/clustername/observeinfraid.go
```diff
@@ -0,0 +1,40 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/configobservation"
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+// ObserveInfraID fills in the cluster-name extended argument for the controller-manager with the cluster's infra ID
+func ObserveInfraID(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
```
deads2k: needs unit test

pkg/operator/configobservation/clustername/observeinfraid.go
```diff
@@ -0,0 +1,40 @@
...
+	infrastructure, err := listers.InfrastructureLister().Get("cluster")
+	if errors.IsNotFound(err) {
+		recorder.Warningf("ObserveInfraID", "Required infrastructures.%s/cluster not found", configv1.GroupName)
+		return observedConfig, errs
+	}
+	if err != nil {
+		return previouslyObservedConfig, errs
+	}
+	infraID := infrastructure.Status.InfrastructureName
+	if err := unstructured.SetNestedStringSlice(observedConfig, []string{infraID}, clusterNamePath...); err != nil {
```
deads2k: only set if you got a value longer than zero.

pkg/operator/configobservation/clustername/observeinfraid.go
```diff
@@ -0,0 +1,40 @@
...
+	observedConfig := map[string]interface{}{}
+	infrastructure, err := listers.InfrastructureLister().Get("cluster")
+	if errors.IsNotFound(err) {
+		recorder.Warningf("ObserveInfraID", "Required infrastructures.%s/cluster not found", configv1.GroupName)
+		return observedConfig, errs
+	}
+	if err != nil {
+		return previouslyObservedConfig, errs
+	}
+	infraID := infrastructure.Status.InfrastructureName
```
deads2k: please add a comment explaining what will happen on upgrade cases.  Is this value always present?

abhinavdahiya: https://github.com/openshift/api/commit/c9b4e5b2d7850deea6719b30aaf38159871ed654 is always set. it was part of the API with 4.0

csrwng:
@deads2k thx for the review, comments addressed

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -119,6 +129,40 @@ func (c TargetConfigController) sync() error {
...
```
deads2k: @sanchezl if this is in two repos now, find a way to put in library-go please

csrwng:
@deads2k so sorry, I realized I was only running one of the tests in the unit test file, (see force-push diff). Can I get another lgtm?

PR #270, opened 2019-07-23
bindata/bootkube: use loopback kubeconfig to talk to API

jhixson74:
This code modifies cluster-kube-controller-manager-operator to use a kubeconfig configured for localhost API access. 

This is necessary due to a limitation with Azure internal load balancers. See limitation #2 here: https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-overview#limitations

"Unlike public Load Balancers which provide outbound connections when transitioning from private IP addresses inside the virtual network to public IP addresses, internal Load Balancers do not translate outbound originated connections to the frontend of an internal Load Balancer as both are in private IP address space. This avoids potential for SNAT port exhaustion inside unique internal IP address space where translation is not required. The side effect is that if an outbound flow from a VM in the backend pool attempts a flow to frontend of the internal Load Balancer in which pool it resides and is mapped back to itself, both legs of the flow don't match and the flow will fail."

kubeconfig-loopback is generated by the installer.

https://jira.coreos.com/browse/CORS-1094

### DIFF
--- a/bindata/bootkube/bootstrap-manifests/kube-controller-manager-pod.yaml
+++ b/bindata/bootkube/bootstrap-manifests/kube-controller-manager-pod.yaml
@@ -19,7 +19,7 @@ spec:
       - containerPort: 10257
     command: ["/bin/bash", "-c"]
     args:
-    - exec hyperkube kube-controller-manager --openshift-config=/etc/kubernetes/config/{{ .ConfigFileName }} --kubeconfig=/etc/kubernetes/secrets/kubeconfig
+    - exec hyperkube kube-controller-manager --openshift-config=/etc/kubernetes/config/{{ .ConfigFileName }} --kubeconfig=/etc/kubernetes/secrets/kubeconfig-loopback
     volumeMounts:
     - mountPath: /etc/ssl/certs
       name: ssl-certs-host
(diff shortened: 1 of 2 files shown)


sttts:
Why is it bad to talk to the bootstrap note during bootstrapping? Why should these control plane operators be different than any other client in the cluster?

sttts:
This misses the kubeconfig in bindata/bootkube/bootstrap-manifests/kube-controller-manager-pod.yaml

jhixson74:
@sttts I have updated the PR description to reflect the problem. I've also updated the pod yaml file. Thanks for catching that.

abhinavdahiya:
So the goal is to make sure the bootstrap-control-plane is using local host to talk to the apisever.
https://github.com/openshift/cluster-kube-controller-manager-operator/pull/270
https://github.com/openshift/cluster-kube-scheduler-operator/pull/156

So rather than changing the kubeconfig useg by the cluster-bootstrap, what if we copy the loopback one to bootstrap-secrets so that bootstrap-{kcm,ks} are talking to the bootstrap-apiserver on localhost..?

or do you think we should use the env variables like https://github.com/openshift/cluster-version-operator/blob/210264aea17b9d6e278435a371775c8eb507568e/bootstrap/bootstrap-pod.yaml#L30-L33 for bootstrap-{kcm,ks} are talking to the bootstrap-apiserver on localhost..?

cc @sttts @jhixson74

sttts:
Lgtm. Why are tests red?

jhixson74:
It looks like this PR is not necessary now so I am closing it.

PR #271, opened 2019-07-26
bump library-go for updated cloudprovider observer

abhinavdahiya:
https://github.com/openshift/library-go/pull/489


abhinavdahiya:
rebased around https://github.com/openshift/cluster-kube-controller-manager-operator/pull/272

@deads2k PTAL

PR #277, opened 2019-08-28
Add trusted-ca configmap for proxy

mfojtik:


mfojtik:
We don't need proxy in this operator/operand yet.

PR #284, opened 2019-09-19
add non-MITM proxy support

deads2k:
Equivalent of https://github.com/openshift/cluster-kube-apiserver-operator/pull/530 for kube-controller-manager.  It's probably worth it to de-dupe the envvar creation itself.

PR #285, opened 2019-09-25
add non-MITM proxy support

soltysh:
This builds on top of #284 but fixes wrong yaml import which was causing the error pointed out by Standa and modifies the configmap merging. 

/assign @deads2k @stlaz

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -10,7 +10,8 @@ import (
 	"strings"
 	"time"
 
-	yaml "gopkg.in/yaml.v2"
+	"github.com/ghodss/yaml"
+
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -536,3 +561,18 @@ func (c *TargetConfigController) namespaceEventHandler() cache.ResourceEventHand
 		},
 	}
 }
+
+func proxyMapToEnvVars(proxyConfig map[string]string) []corev1.EnvVar {
```
mfojtik: TODO: move this to library-go :-)

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -291,6 +302,20 @@ func managePod(configMapsGetter corev1client.ConfigMapsGetter, secretsGetter cor
...
+	if err := yaml.Unmarshal(operatorSpec.ObservedConfig.Raw, &observedConfig); err != nil {
+		return nil, false, fmt.Errorf("failed to unmarshal the observedConfig: %v", err)
+	}
+	proxyConfig, _, err := unstructured.NestedStringMap(observedConfig, "targetconfigcontroller", "proxy")
+	if err != nil {
+		return nil, false, fmt.Errorf("couldn't get the proxy config from observedConfig: %v", err)
+	}
+
+	proxyEnvVars := proxyMapToEnvVars(proxyConfig)
+	for i, container := range required.Spec.Containers {
```
mfojtik: how about initContainers?

soltysh: We can add it to 4.3

PR #290, opened 2019-10-01
Bump to kubernetes 1.16

soltysh:
/assign @tnozicka @sallyom


tnozicka:
/lgtm

/hold
(wanna sync fist if we shouldn't use gomod to get correct transitive dependency levels, or we do it as follow up)

soltysh:
> (wanna sync fist if we shouldn't use gomod to get correct transitive dependency levels, or we do it as follow up)

followup
/hold cancel

PR #291, opened 2019-10-03
Update manifests with new crd-schema-gen from library-go

damemi:
This bumps library-go to pull in the new crd-schema-gen based on upstream controllers tools and regenerates CRDs to account for structural schema requirements in 1.16


damemi:
/hold
making changes to the library-go generator

damemi:
These changes were brought in by https://github.com/openshift/cluster-kube-controller-manager-operator/pull/292

PR #300, opened 2019-10-24
Bug 1767047: move rbac for cluster-policy-controller leader lock to kcm-o

sallyom:
this is a pick of 425dfc3970149b6f1d28d

/cc @soltysh


derekwaynecarr:
The referenced bug was moved to VERIFIED by dev.  This messes up our process, and has been updated to ensure QE also verifies the bug.

/bugzilla refresh

PR #301, opened 2019-10-31
Bug 1762888: alert when number of expected pods is equal or lower than desired healthy pods in a PDB

soltysh:
/assign @smarterclayton @eparis @sferich888 @cuppett

### DIFF
--- a/manifests/0000_90_kube-controller-manager-operator_03_servicemonitor.yaml
+++ b/manifests/0000_90_kube-controller-manager-operator_03_servicemonitor.yaml
@@ -42,3 +42,29 @@ spec:
       for: 15m
       labels:
         severity: critical
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: kube-controller-manager-operator
+  namespace: openshift-kube-controller-manager-operator
+spec:
+  groups:
+  - name: cluster-version
+    rules:
+    - alert: PodDisruptionBudgetAtLimit
+      annotations:
+        message: The pod disruption budget is preventing further disruption to pods because it is at the minimum allowed level.
+      expr: |
+        kube_poddisruptionbudget_status_expected_pods == on(poddisruptionbudget, service) kube_poddisruptionbudget_status_desired_healthy
+      for: 15m
+      labels:
+        severity: warning
+    - alert: PodDisruptionBudgetLimit
+      annotations:
+        message: The pod disruption budget is below the minimum number allowed pods.
+      expr: |
+        kube_poddisruptionbudget_status_expected_pods < on(poddisruptionbudget, service) kube_poddisruptionbudget_status_desired_healthy
+      for: 15m
+      labels:
+        severity: critical


manifests/0000_90_kube-controller-manager-operator_03_servicemonitor.yaml
```diff
@@ -42,3 +42,21 @@ spec:
...
+  namespace: openshift-kube-controller-manager-operator
+spec:
+  groups:
+  - name: cluster-version
+    rules:
+    - alert: KubeControllerManagerDown
+      annotations:
+        message: KubeControllerManager has disappeared from Prometheus target discovery.
+      expr: |
+        kube_poddisruptionbudget_status_expected_pods <= on(poddisruptionbudget, service) kube_poddisruptionbudget_status_desired_healthy
```
soltysh: This should print the PDB and its service when expected pods will less or equal than desired healthy number.

manifests/0000_90_kube-controller-manager-operator_03_servicemonitor.yaml
```diff
@@ -42,3 +42,21 @@ spec:
...
+      for: 15m
+      labels:
+        severity: warning
```
soltysh: I spoke with monitoring team, we don't have info level, so I went with a warning. I'm open to suggestions, though.

manifests/0000_90_kube-controller-manager-operator_03_servicemonitor.yaml
wking: > ...less or equal than desired healthy number...

Is this what we want?  We don't want to alert when you drop <= `minAvailable` or some such?  Does this fire during upgrade when we cycle a control plane machine and at least while that is happening should be one less than our desired count but still not violating our disruption budget.

wking: Pointed [PromeCIeus][1] at [the CI upgrade run][2], looks like this is fine:

```
(max(kube_poddisruptionbudget_status_expected_pods) by (poddisruptionbudget, service)) - (min (kube_poddisruptionbudget_status_desired_healthy) by (poddisruptionbudget, service))
```

Screenshot from 2019-11-06 12-28-04

For example, `kube_poddisruptionbudget_status_expected_pods` is 3 for etcd-quorum-guard over that whole time range, while `kube_poddisruptionbudget_status_desired_healthy` is 2.

[1]: https://promecieus.svc.ci.openshift.org/
[2]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-kube-controller-manager-operator/301/pull-ci-openshift-cluster-kube-controller-manager-operator-master-e2e-aws-upgrade/209

manifests/0000_90_kube-controller-manager-operator_03_servicemonitor.yaml
smarterclayton: Wait.... I know there are info level things.  I swear I created one...

manifests/0000_90_kube-controller-manager-operator_03_servicemonitor.yaml
```diff
@@ -42,3 +42,21 @@ spec:
...
+spec:
+  groups:
+  - name: cluster-version
+    rules:
+    - alert: KubeControllerManagerDown
+      annotations:
+        message: KubeControllerManager has disappeared from Prometheus target discovery.
+      expr: |
+        kube_poddisruptionbudget_status_expected_pods <= on(poddisruptionbudget, service) kube_poddisruptionbudget_status_desired_healthy
+      for: 15m
```
smarterclayton: This seems pretty reasonable, how does this compare to others.

manifests/0000_90_kube-controller-manager-operator_03_servicemonitor.yaml
soltysh: Where, I've been looking for them and monitoring team explicitly told me there none.

manifests/0000_90_kube-controller-manager-operator_03_servicemonitor.yaml
soltysh: It's similar to the others, even though the bug mentioned 5 min, I went with what the other alerts usually have.

smarterclayton: That's fine for me.

manifests/0000_90_kube-controller-manager-operator_03_servicemonitor.yaml
smarterclayton: Hrm, I guess I was wrong.  This is fine, and is definitely an alert.

manifests/0000_90_kube-controller-manager-operator_03_servicemonitor.yaml
```diff
@@ -42,3 +42,21 @@ spec:
...
+metadata:
+  name: kube-controller-manager-operator
+  namespace: openshift-kube-controller-manager-operator
+spec:
+  groups:
+  - name: cluster-version
+    rules:
+    - alert: KubeControllerManagerDown
+      annotations:
+        message: KubeControllerManager has disappeared from Prometheus target discovery.
```
smarterclayton: I think these messages are wrong?

manifests/0000_90_kube-controller-manager-operator_03_servicemonitor.yaml
```diff
@@ -42,3 +42,21 @@ spec:
...
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: kube-controller-manager-operator
+  namespace: openshift-kube-controller-manager-operator
+spec:
+  groups:
+  - name: cluster-version
+    rules:
+    - alert: KubeControllerManagerDown
```
smarterclayton: Isn't this the wrong name?

manifests/0000_90_kube-controller-manager-operator_03_servicemonitor.yaml
```diff
@@ -42,3 +42,21 @@ spec:
...
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: kube-controller-manager-operator
+  namespace: openshift-kube-controller-manager-operator
+spec:
+  groups:
+  - name: cluster-version
+    rules:
+    - alert: PodDisruptionBudgetMisconfigured
```
smarterclayton: PodDisruptionBudgetPreventsDrain

smarterclayton: Actually..

PodDisruptionBudgetAtLimit = PDB expected == health: "The pod disruption budget is preventing further disruption to pods because it is at the minimum allowed level" - warning
PodsBelowDisruptionLimit = PDB expected < health: "The pod disruption budget is below the minimum number allowed pods" - critical

smarterclayton:
/lgtm

LOVE IT

soltysh:
@smarterclayton since ^ failed, I've opened https://github.com/openshift/cluster-kube-controller-manager-operator/pull/306

smarterclayton:
When reviewing this, I don't see the namespace of the PDB reported in the alert, only the name.  I'm not sure if this is a bug in the alert on in how the PDB metric is generated.

PR #302, opened 2019-11-04
Bump library go

soltysh:
/assign @sttts @damemi 
This brings updated library-go (I care here for the events for cert rotation needed for autorecovery) and switches to using vendored CRD definition.


damemi:
crd changes look good to me
/lgtm

PR #306, opened 2019-11-08
Bug 1770303: alert when number of expected pods is equal or lower than desired healthy pods in a PDB

soltysh:
/assign @smarterclayton


soltysh:
Also includes https://github.com/openshift/cluster-kube-controller-manager-operator/pull/309

soltysh:
>     * expected dependent [Bugzilla bug 1762888](https://bugzilla.redhat.com/show_bug.cgi?id=1762888) to target the "4.3.z" release, but it targets "4.3.0" instead

I'm going to manually change the valid bug label.

PR #308, opened 2019-11-12
Fix metrics

mfojtik:
* First commit will make service monitor secure and split the alerts to separate files (lists are evil in CVO)
* Second commit will fix collection of operator version metrics


manifests/0000_90_kube-controller-manager-operator_03_servicemonitor.yaml
```diff
@@ -24,47 +24,4 @@ spec:
   selector:
     matchLabels:
       app: kube-controller-manager-operator
----
```
sttts: moves in their own commits please.

smarterclayton:
This is going to 4.2 as well (fixes cluster_infrastructure_provider?)?

PR #309, opened 2019-11-12
Add namespace to PDB limit alert

soltysh:
/assign @smarterclayton

### DIFF
--- a/manifests/0000_90_kube-controller-manager-operator_05_alert-pdb.yaml
+++ b/manifests/0000_90_kube-controller-manager-operator_05_alert-pdb.yaml
@@ -13,7 +13,7 @@ spec:
           annotations:
             message: The pod disruption budget is preventing further disruption to pods because it is at the minimum allowed level.
           expr: |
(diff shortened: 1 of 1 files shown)


smarterclayton:
/hold

You might have to get review from monitoring here, the "namespace" label is kind of special (it might be automagic)

smarterclayton:
@s-urbaniak is `namespace` the correct label (I assume it gets renamed to `exported_namespace`)?

s-urbaniak:
(at least for this metric)

LiliC:
Yes the namespace is there and set as the `PodDisruptionBudget` resource ns. 👍 

[source code](https://github.com/kubernetes/kube-state-metrics/blob/e03540f06bfd3a933b2178c975f1f9d1a382ac96/internal/store/poddisruptionbudget.go#L133)

soltysh:
Rebased, after https://github.com/openshift/cluster-kube-controller-manager-operator/pull/308 landed, I will self-tag since this is a simple fix already approved.

PR #311, opened 2019-11-12
Bug 1744984: Add e2e test for kube controller manager lease configmap deletion

damemi:
This is a work in progress, trying to create a simple e2e that deletes the KCM configmap then attempts to create a replicaset to see if KCM recovered from the deletion, as described here: https://bugzilla.redhat.com/show_bug.cgi?id=1744984

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -2,8 +2,11 @@ package e2e
 
 import (
 	"testing"
+	"time"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/operatorclient"
@@ -20,6 +23,40 @@ func TestOperatorNamespace(t *testing.T) {
 		t.Fatal(err)
 	}
 	_, err = kubeClient.CoreV1().Namespaces().Get(operatorclient.OperatorNamespace, metav1.GetOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestKCMRecovery(t *testing.T) {
+	// This is an e2e test to verify that KCM can recover from having its lease configmap deleted
+	// See https://bugzilla.redhat.com/show_bug.cgi?id=1744984
+	kubeConfig, err := test.NewClientConfigForTest()
+	if err != nil {
+		t.Fatal(err)
+	}
+	kubeClient, err := kubernetes.NewForConfig(kubeConfig)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Try to delete the kube controller manager's configmap in kube-system
+	err = kubeClient.CoreV1().ConfigMaps("kube-system").Delete("kube-controller-manager", &metav1.DeleteOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Check to see that the configmap then gets recreated
+	err = wait.Poll(time.Second*5, time.Second*300, func() (bool, error) {
+		_, err := kubeClient.CoreV1().ConfigMaps("kube-system").Get("kube-controller-manager", metav1.GetOptions{})
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+		if err != nil {
+			return false, err
+		}
+		return true, nil
+	})
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/bindata/v4.1.0/kube-controller-manager/leader-election-kube-controller-manager-role-kube-system.yaml
+++ b/bindata/v4.1.0/kube-controller-manager/leader-election-kube-controller-manager-role-kube-system.yaml
@@ -1,0 +1,17 @@
+# This role is necessary to create leader lock configmap for upgrades 4.2-> 4.3
+# cluster-policy-controller is split from openshift-controller-manager in 4.3
+# leader lock in openshift-controller-manager NamespaceSecurityAllocationController and in ClusterPolicyController
+# cluster-policy-controller container runs in ns openshift-kube-controller-manager static pod
+# The lock, role, and rolebinding can be removed in 4.4
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: kube-system
+  name: system:openshift:leader-election-lock-kube-controller-manager
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - create
--- a/bindata/v4.1.0/kube-controller-manager/leader-election-kube-controller-manager-rolebinding-kube-system.yaml
+++ b/bindata/v4.1.0/kube-controller-manager/leader-election-kube-controller-manager-rolebinding-kube-system.yaml
@@ -1,0 +1,19 @@
+# This rolebinding binds role for creation of leader lock configmap for upgrades 4.2-> 4.3
+# cluster-policy-controller is split from openshift-controller-manager in 4.3
+# leader lock in openshift-controller-manager NamespaceSecurityAllocationController and in ClusterPolicyController
+# cluster-policy-controller container runs in ns openshift-kube-controller-manager static pod
+# The locks, role, and rolebinding can be removed in 4.4
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  namespace: kube-system
+  name: system:openshift:leader-election-lock-kube-controller-manager
+roleRef:
+  kind: Role
+  name: system:openshift:leader-election-lock-kube-controller-manager
+subjects:
+- kind: User
+  name: system:kube-controller-manager
+- kind: ServiceAccount
+  name: namespace-security-allocation-controller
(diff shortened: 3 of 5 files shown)


test/e2e/operator_test.go
```diff
@@ -24,3 +26,68 @@ func TestOperatorNamespace(t *testing.T) {
...
+						},
+					},
+				},
+			},
+		},
+	}
+	_, err = kubeClient.AppsV1().ReplicaSets(operatorclient.OperatorNamespace).Create(replicaSet)
+	if err != nil {
+		t.Fatal(err)
+	}
```
soltysh: You're missing a bit where you wait and verify if that replica set actually run and KCM recovered.

damemi: Updated to add a wait loop that looks for pods created with the replicaset labels, that would indicate the KCM has recovered right?

test/e2e/operator_test.go
```diff
@@ -24,3 +28,80 @@ func TestOperatorNamespace(t *testing.T) {
...
+	for _, pod := range pods.Items {
+		err := kubeClient.CoreV1().Pods("openshift-kube-controller-manager").Delete(pod.Name, &metav1.DeleteOptions{})
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// Now try to create a replicaset and see that it is successfully created by the controller manager
```
deads2k: wait for the original pods to be gone before creating this.

test/e2e/operator_test.go
```diff
@@ -24,3 +28,80 @@ func TestOperatorNamespace(t *testing.T) {
...
+	}
+
+	// Try to delete the kube controller manager's configmap in kube-system
+	err = kubeClient.CoreV1().ConfigMaps("kube-system").Delete("kube-controller-manager", &metav1.DeleteOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Get all the currently running KCM pods, and then delete them
+	pods, err := kubeClient.CoreV1().Pods("openshift-kube-controller-manager").List(metav1.ListOptions{})
```
deads2k: this is pointless

test/e2e/operator_test.go
```diff
@@ -24,3 +28,80 @@ func TestOperatorNamespace(t *testing.T) {
...
+							Name:    "test",
+							Image:   "busybox:latest",
+							Command: []string{"/bin/true"},
+						},
+					},
+				},
+			},
+		},
+	}
+	_, err = kubeClient.AppsV1().ReplicaSets(operatorclient.OperatorNamespace).Create(replicaSet)
```
deads2k: because the pods aren't gone, this doesn't matter

test/e2e/operator_test.go
```diff
@@ -24,3 +28,80 @@ func TestOperatorNamespace(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Poll to see if pods from replicaset are created which would indicate KCM recovered
+	wait.Poll(time.Second*5, time.Second*300, func() (bool, error) {
```
deads2k: remove this

test/e2e/operator_test.go
```diff
@@ -24,3 +28,80 @@ func TestOperatorNamespace(t *testing.T) {
...
+		pods, err := kubeClient.CoreV1().Pods(operatorclient.OperatorNamespace).List(metav1.ListOptions{LabelSelector: "app=test"})
+		if err != nil {
+			return false, err
+		}
+		if len(pods.Items) > 0 {
+			return true, nil
+		}
+		return false, nil
+	})
+}
```
deads2k: add a check that the configmap gets recreated

test/e2e/operator_test.go
```diff
@@ -24,3 +27,34 @@ func TestOperatorNamespace(t *testing.T) {
...
+	}
+
+	// Try to delete the kube controller manager's configmap in kube-system
+	err = kubeClient.CoreV1().ConfigMaps("kube-system").Delete("kube-controller-manager", &metav1.DeleteOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Check to see that the configmap then gets recreated
+	wait.Poll(time.Second*5, time.Second*300, func() (bool, error) {
```
deads2k: you forgot to check this error.

test/e2e/operator_test.go
```diff
@@ -24,3 +27,34 @@ func TestOperatorNamespace(t *testing.T) {
...
+		_, err := kubeClient.CoreV1().ConfigMaps("kube-system").Get("kube-controller-manager", metav1.GetOptions{})
+		if err != nil {
```
deads2k:
```go
if errors.IsNotFound(err){
    return false, nil
}
if err != nil{
    return false, err
}
return true, nil
```

soltysh:
/hold
@damemi it looks like the test keeps failing

deads2k:
> /hold
> @damemi it looks like the test keeps failing

hehe.  I know you didn't doubt me :)

damemi:
Looking more into this, it seems we use 2 roles: `system:openshift:leader-election-lock-cluster-policy-controller` in the `openshift-kube-controller-manager` namespace (see [role](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/bindata/v4.1.0/kube-controller-manager/leader-election-cluster-policy-controller-role.yaml), [binding](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/bindata/v4.1.0/kube-controller-manager/leader-election-cluster-policy-controller-rolebinding.yaml)) and `system::leader-locking-kube-controller-manager` for the `kube-system` namespace. However, the latter [does not include permissions to create configmaps](https://github.com/kubernetes/kubernetes/blob/292bd0efea53a1416a6a7d01620779d2d72fec96/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/namespace_policy.go#L107-L114), so I'm attempting to change our `openshift` role to a `ClusterRole`, and adding a second binding for it in the `kube-system` namespace

damemi:
Spun up a cluster based on this PR, and it doesn't look like the new rolebinding I added in `kube-system` is created... digging more

damemi:
More info: the `kube-system` rolebinding isn't created, but the (now) ClusterRole is. When I manually create the system rolebinding to that clusterrole and `oc delete cm/kube-controller-manager -n kube-system`, I see the configmap almost instantly recreated. So, if I can just figure out how to get that system rolebinding to be automatically created this should fix the bug and pass the test

damemi:
Splitting the rolebindings into 2 separate files got the operator e2e to pass (maybe resourceapply doesn't do well with a list of resources to parse). Squashing and hoping everything goes green again

damemi:
This is failing upgrade test with the following error:
```
ns/openshift-kube-controller-manager-operator deployment/kube-controller-manager-operator Failed to update RoleBinding.rbac.authorization.k8s.io/system:openshift:leader-election-lock-cluster-policy-controller -n openshift-kube-controller-manager: RoleBinding.rbac.authorization.k8s.io "system:openshift:leader-election-lock-cluster-policy-controller" is invalid: roleRef: Invalid value: rbac.RoleRef{APIGroup:"rbac.authorization.k8s.io", Kind:"ClusterRole", Name:"system:openshift:leader-election-lock-cluster-policy-controller"}: cannot change roleRef (798 times)
```
@deads2k do you have any idea why the `roleRef` would be valid for the other tests, but fail on an upgrade?

soltysh:
> @deads2k do you have any idea why the `roleRef` would be valid for the other tests, but fail on an upgrade?

4.3 is where we split the cluster-policy-controller, that's where this problem might be coming from

damemi:
Ok, switching to 2 namespaced `Roles` (as opposed to 1 `ClusterRole`) seems to fix the problem, and all tests pass. Squashing now and this is ready for final review
cc @soltysh @deads2k 
/bugzilla refresh

bindata/v4.1.0/kube-controller-manager/leader-election-cluster-policy-controller-role-kube-system.yaml
```diff
@@ -0,0 +1,22 @@
+# This role is necessary to create leader lock configmap for upgrades 4.2-> 4.3
+# cluster-policy-controller is split from openshift-controller-manager in 4.3
+# leader lock in openshift-controller-manager NamespaceSecurityAllocationController and in ClusterPolicyController
+# cluster-policy-controller container runs in ns openshift-kube-controller-manager static pod
+# The lock, role, and rolebinding can be removed in 4.4
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: kube-system
+  name: system:openshift:leader-election-lock-cluster-policy-controller
```
deads2k: this name doesn't match what you're binding to.  This is a general leader election role

bindata/v4.1.0/kube-controller-manager/leader-election-cluster-policy-controller-role-kube-system.yaml
```diff
@@ -0,0 +1,22 @@
...
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - watch
+  - create
```
deads2k: only create needs to be unconstrained by names. The rest can use the stock role, right?

damemi:
@deads2k updated with your feedback, thanks

PR #312, opened 2019-11-13
Add manifest annotations for hosted deployment exclusions

csrwng:
Enables the CVO to exclude manifests in an externally hosted control plane deployment
See https://github.com/openshift/cluster-version-operator/pull/252


csrwng:
/assign @tnozicka 
This should be ready for merge

PR #315, opened 2019-11-15
Bug 1758963: hide secrets in logs

soltysh:
/assign @tnozicka


soltysh:
Manually adding approved, since I'm not in the 4.1 owners

PR #316, opened 2019-11-20
Bug 1771809: Add e2e for PodDisruptionBudgetAtLimit alert

sallyom:
Here's an e2e that:
* creates a pod
* creates a PodDisruptionBudget  of minAvail=1
* queries prometheus for a pending 'PodDisruptionBudgetAtLimit' alert 'pending' - to wait for 'firing' you'd have to wait 15 min, so pending it is. 
* query uses helper functions from this PR: https://github.com/openshift/library-go/pull/602 (fake bump here) 

/cc @soltysh @marun


test/e2e/operator_test.go
```diff
@@ -1,26 +1,316 @@
...
+	"github.com/prometheus/common/model"
 )
 
-func TestOperatorNamespace(t *testing.T) {
+const (
+	kcmoNamespace        = "openshift-kube-controller-manager-operator"
+	kcmoSigningKeySecret = "kube-controller-manager-operator-serving-cert"
+)
+
+func TestE2E(t *testing.T) {
```
sanchezl: I'm not sure this nesting is needed. All tests under `test/e2e` are e2e tests, and for example , `TestSATokenSignerControllerSyncCerts` is in another file and won't be in this hierarchy.

marun: Nesting tests allows setup in the enclosing function to be shared, but I would agree that unless the setup is expensive (e.g. creating resources) I'm not sure it makes sense to use nesting vs creating a helper that tests can call.

marun: @sanchezl Ah, and to your point, naming this `TestE2E` probably doesn't make sense unless this is the entrypoint for running all e2e tests rather than just some of them.

test/e2e/operator_test.go
```diff
@@ -1,26 +1,316 @@
...
 	}
-	_, err = kubeClient.CoreV1().Namespaces().Get(operatorclient.OperatorNamespace, metav1.GetOptions{})
+
+	policyClient, err := policyclientv1beta1.NewForConfig(kubeConfig)
 	if err != nil {
 		t.Fatal(err)
 	}
+
+	// test that operator namespace exists
+	t.Run("OperatorNamespace", func(t *testing.T) {
```
marun: Is the operator namespace being present a precondition for the other test to execute? Since `t.Run` isolates failure, enclosing this check would result in suite failure but would not prevent `PodDisruptionBudgetAtLimit alert` from executing.

test/e2e/operator_test.go
```diff
@@ -1,26 +1,316 @@
...
+		_, err = kubeClient.CoreV1().Namespaces().Get(operatorclient.OperatorNamespace, metav1.GetOptions{})
+		if err != nil {
+			t.Fatal(err)
+		}
+	})
+
+	// test that pdb-atlimit alert exists when there is a pdb at limit
+	t.Run("PodDisruptionBudgetAtLimit alert", func(t *testing.T) {
```
marun: (No action required) Consider using of dashes or underscores in test names instead of spaces to simplify targeting test execution.

test/e2e/operator_test.go
```diff
@@ -1,26 +1,316 @@
...
+		},
+	})
+	return ns, err
+}
+
+func cleanupTestObjects(client *kubernetes.Clientset, ns string) {
+	client.CoreV1().Namespaces().Delete(ns, &metav1.DeleteOptions{})
+}
+
+// took this from service-ca-operator test
```
marun: Suggesting using uuid instead:

https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/vendor/github.com/openshift/library-go/pkg/config/leaderelection/leaderelection.go#L32

marun: But this isn't really the best way to deal with ensuring name uniqueness in tests. Setting `ObjectMeta.GenerateName` is likely preferable if it's possible to create the object before knowing the name:

https://github.com/kubernetes-sigs/kubefed/blob/master/test/e2e/framework/framework.go#L308

It's also possible to directly call the apiserver's name generating logic:

https://github.com/kubernetes-sigs/kubefed/blob/master/test/e2e/defaulting.go#L66

test/e2e/operator_test.go
```diff
@@ -1,26 +1,316 @@
...
+
+type prometheusResponseData struct {
+	ResultType string       `json:"resultType"`
+	Result     model.Vector `json:"result"`
+}
+
+func getBearerTokenURLViaPod(t *testing.T, kclient *kubernetes.Clientset, ns, execPodName, url, bearer string) ([]byte, error) {
+	//cmd := fmt.Sprintf("curl -s -k -H 'Authorization: Bearer %s' %q", bearer, url)
+	// Retrieve the cert and key PEM of the current CA via kube-controller-manager-operator serving-cert secret
+	secret, err := kclient.CoreV1().Secrets(kcmoNamespace).Get(kcmoSigningKeySecret, metav1.GetOptions{})
```
marun: That configuring tls with the serving cert works is an accident of implementation. The signing CA is included with the serving cert, allowing it to be reused as a ca bundle. However, this is not good practice and at some point in the future serving certs will not include the signing CA. Please source an actual ca bundle from the service ca operator instead as per the following example (create annotated configmap, wait ca bundle to be added to configmap):

https://github.com/openshift/service-ca-operator/blob/ca-validity-duration/test/e2e/e2e_test.go#L329

test/e2e/operator_test.go
```diff
@@ -1,26 +1,316 @@
...
+	return err
+}
+
+func locatePrometheus(t *testing.T, client *kubernetes.Clientset) (url, bearerToken string, ok bool) {
+	_, err := client.CoreV1().Services("openshift-monitoring").Get("prometheus-k8s", metav1.GetOptions{})
+	if kapierrs.IsNotFound(err) {
+		return "", "", false
+	}
+
+	for i := 0; i < 30; i++ {
```
marun: Please use one of the [wait.Poll](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go#L333) functions instead of a manual loop e.g.

https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/vendor/github.com/openshift/library-go/pkg/operator/certrotation/config.go#L17

test/e2e/operator_test.go
```diff
@@ -1,26 +1,316 @@
...
+			if err != nil {
+				return false, fmt.Errorf("Waiting for prometheus: %v", err)
+			} else {
+				return true, nil
+			}
+		})
+	})
+}
+
+func pdbCreate(t *testing.T, client *policyclientv1beta1.PolicyV1beta1Client, name string, labels map[string]string) error {
```
marun: `t` is supplied to this method and the one that follows but does not appear to be used (maybe a linter is in order).

test/e2e/operator_test.go
```diff
@@ -1,26 +1,316 @@
...
+func podCreate(t *testing.T, client *kubernetes.Clientset, name string, labels map[string]string) error {
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-pod",
+			Namespace: name,
+			Labels:    labels,
+		},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{Name: "test", Image: "centos:7", Command: []string{"sh", "-c", "trap exit TERM; while true; do sleep 5; done"}},
```
marun: Please prefer breaking long lines on parameters for readability, e.g

```golang
{
  Name: "test", 
  Image: "centos:7", 
  Command: []string{
     "sh", 
     "-c", 
     "trap exit TERM; while true; do sleep 5; done",
  },
}
```

test/e2e/operator_test.go
```diff
@@ -1,26 +1,316 @@
...
+			}
+			return false, nil
+		})
+
+		err = pdbCreate(t, policyClient, name, labels)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		fmt.Println("Waiting 15s for PodDisruptionBudget to process")
```
marun: Why is it desirable to sleep for 15 seconds vs just relying on the subsequent polling loop?

test/e2e/operator_test.go
```diff
@@ -1,26 +1,316 @@
...
+		}
+		defer cleanupTestObjects(kubeClient, ns.Name)
+
+		err = podCreate(t, kubeClient, name, labels)
+		if err != nil {
+			t.Fatal(err)
+		}
+		var pods *corev1.PodList
+		// Poll to see if pod is created
+		wait.Poll(time.Second*5, time.Second*300, func() (bool, error) {
```
marun: If line 71 executed successfully, what would cause the pod to disappear?

test/e2e/operator_test.go
```diff
@@ -1,26 +1,316 @@
...
+
+	body, err := ioutil.ReadAll(resp.Body)
+	return body, nil
+}
+
+// runQuery is similar to runQueries that lives in origin/test/extended/prometheus
+func runQuery(t *testing.T, kclient *kubernetes.Clientset, query, ns, execPodName, baseURL, bearerToken string, expected bool) error {
+	// expect all correct metrics within a reasonable time period
+	var err error
+	for i := 0; i < waitForPrometheusStartSeconds; i++ {
```
marun: Please prefer a `wait.Poll` method.

test/e2e/operator_test.go
sallyom: Good point, I removed the nesting.

test/e2e/operator_test.go
sallyom: updated, took out the nesting.

test/e2e/operator_test.go
sallyom: looking into it...

test/e2e/operator_test.go
sallyom: yes, correct, cleaning that up

test/e2e/operator_test.go
sallyom: will do

test/e2e/operator_test.go
sallyom: ok, got it.. working on it...

test/e2e/operator_test.go
sallyom: I've added a wait.Poll for this

test/e2e/operator_test.go
sallyom: ok, done

test/e2e/operator_test.go
sallyom: wait.Poll for "Running" (I added that), not sure it's necessary, though

test/e2e/operator_test.go
sallyom: added wait.Poll there

test/e2e/operator_test.go
sallyom: ah, I knew there must be a better way.  fixed.

test/e2e/operator_test.go
sallyom: done, i'm obtaining the caBundle

sallyom:
@marun @sanchezl thanks for the review! I've addressed your suggestions, let me know if this (or more generic obtain caBundle/query functions, at least) should be moved to library-go.

go.mod
```diff
@@ -36,7 +37,7 @@ replace (
 	github.com/jteeuwen/go-bindata => github.com/jteeuwen/go-bindata v3.0.8-0.20151023091102-a0ff2567cfb7+incompatible
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20191101174058-d428681bd3ce
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20191022152013-2823239d2298
-	github.com/openshift/library-go => github.com/openshift/library-go v0.0.0-20191121124438-7c776f7cc17a
+	github.com/openshift/library-go => github.com/sallyom/library-go v0.0.0-20191125154451-7578d03d01df
```
sttts: only for testing?

test/e2e/operator_test.go
```diff
@@ -24,3 +35,129 @@ func TestOperatorNamespace(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	pq, err := metrics.NewPromQuery(kubeClient, pdbAlertQuery, cmName, ns.Name)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Now check for alert
+	wait.Poll(time.Second*5, time.Second*300, func() (bool, error) {
```
sttts: wait.InfiniteTestTimeout or something like that instead of 5min

sttts: same above

test/e2e/operator_test.go
```diff
@@ -24,3 +35,129 @@ func TestOperatorNamespace(t *testing.T) {
...
+	}
+	var pods *corev1.PodList
+	// Poll to see if pod is created
+	wait.Poll(time.Second*5, time.Second*300, func() (bool, error) {
+		pods, err = kubeClient.CoreV1().Pods(name).List(metav1.ListOptions{LabelSelector: "app=pdbtest"})
+		if err != nil {
+			return false, err
+		}
+		if len(pods.Items) > 0 {
+			if pods.Items[0].Status.Phase == corev1.PodRunning {
```
sttts: combine into one if

test/e2e/operator_test.go
```diff
@@ -24,3 +35,129 @@ func TestOperatorNamespace(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+// TestPodDisruptionBudgetAtLimitAlert tests that pdb-atlimit alert exists when there is a pdb at limit
+func TestPodDisruptionBudgetAtLimitAlert(t *testing.T) {
+	// See https://bugzilla.redhat.com/show_bug.cgi?id=1762888
```
sttts: nit: move into func comment

test/e2e/operator_test.go
```diff
@@ -1,15 +1,26 @@
...
+	"k8s.io/apiserver/pkg/storage/names"
 	"k8s.io/client-go/kubernetes"
+	policyclientv1beta1 "k8s.io/client-go/kubernetes/typed/policy/v1beta1"
 
 	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/operatorclient"
 	test "github.com/openshift/cluster-kube-controller-manager-operator/test/library"
+	metrics "github.com/openshift/library-go/pkg/metrics"
 )
 
+const pdbAlertQuery = `ALERTS{alertstate="pending",alertname="PodDisruptionBudgetAtLimit",severity="warning"} == 1`
```
sttts: would inline this into the func.

go.mod
sallyom: yes, only for testing, will remove once library-go pr is merged.. i'll hold until then.

sallyom:
/hold 

until fake bump is removed, library-go pr merged: https://github.com/openshift/library-go/pull/602

test/e2e/operator_test.go
sallyom: I've moved to `wait.PollImmediate(time.Second*1, testTimeout, func() (bool, error) {`
and testTimeout 120s.  Test time was 25s (oops didn't realize how long that was) down to 14s.

test/e2e/operator_test.go
sallyom: done

test/e2e/operator_test.go
sallyom: done

test/e2e/operator_test.go
sallyom: ok, done

sallyom:
@sttts thanks for the review, i've updated, ready for another look.

test/e2e/operator_test.go
```diff
@@ -24,3 +33,136 @@ func TestOperatorNamespace(t *testing.T) {
...
+		if _, err := policyClient.PodDisruptionBudgets(name).List(metav1.ListOptions{LabelSelector: "app=pbtest"}); err != nil {
+			return false, fmt.Errorf("Waiting for poddisruptionbudget")
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	caBundle, err := metrics.RetrieveServiceCABundle(kubeClient, cmName, ns.Name)
```
marun: I'm happy to see this decomposed implementation - I think it will prove easier to maintain. I think the next step is adding a helper method to simplify calling for cases where only a single query is required e.g.

```golang
metrics.QueryPrometheus(kubeClient, ns.Name, query)
```

test/e2e/operator_test.go
```diff
@@ -24,3 +33,136 @@ func TestOperatorNamespace(t *testing.T) {
...
+	bearerToken, ok := metrics.LocatePrometheus(kubeClient)
+	if !ok {
+		t.Fatalf("Could not locate prometheus")
+	}
+	pq, err := metrics.NewPrometheusClient(caBundle, `ALERTS{alertstate="pending",alertname="PodDisruptionBudgetAtLimit",severity="warning"} == 1`, bearerToken)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Now check for alert
+	wait.PollImmediate(time.Second*1, testTimeout, func() (bool, error) {
```
marun: Is this polling loop necessary given that `RunPrometheusQuery` is already running a polling loop?

sallyom: nope, updated : )

sallyom:
I notice the CI template cluster-launch-installer-src test pod does not have access to oc/kubectl, so will have to work around that or add the cli to the CI template. (that's why the aws-operator test fails here, tests run locally fine as long as oc/kubectl is on the system)

test/e2e/operator_test.go
```diff
@@ -24,3 +38,126 @@ func TestOperatorNamespace(t *testing.T) {
...
+	err = metrics.PodCreate(kubeClient, name, cmName, labels)
+	if err != nil {
+		t.Fatal(err)
+	}
+	var pods *corev1.PodList
+	// Poll to confirm pod is running
+	wait.PollImmediate(time.Second*1, testTimeout, func() (bool, error) {
+		pods, err = kubeClient.CoreV1().Pods(name).List(metav1.ListOptions{LabelSelector: "app=pdbtest"})
+		if err != nil {
+			return false, err
```
marun: Is this error fatal?

Makefile
```diff
@@ -34,6 +34,7 @@ $(call build-image,ocp-cluster-kube-controller-manager-operator,$(IMAGE_REGISTRY
 $(call add-bindata,v4.1.0,./bindata/v4.1.0/...,bindata,v411_00_assets,pkg/operator/v411_00_assets/bindata.go)
 
 test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
+test-e2e: GO_TEST_FLAGS += -v
```
soltysh: Ditto as in the other PR, I don't think we need it here.

test/e2e/operator_test.go
```diff
@@ -24,3 +33,136 @@ func TestOperatorNamespace(t *testing.T) {
...
+
+	err = pdbCreate(policyClient, name, labels)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Wait for pbd
+	err = wait.PollImmediate(time.Second*1, testTimeout, func() (bool, error) {
+		if _, err := policyClient.PodDisruptionBudgets(name).List(metav1.ListOptions{LabelSelector: "app=pbtest"}); err != nil {
+			return false, fmt.Errorf("Waiting for poddisruptionbudget")
```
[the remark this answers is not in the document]
soltysh: Now with go1.13:
```
fmt.Errorf("waiting for poddisruptionbudget: %w", err)
```

test/e2e/operator_test.go
```diff
@@ -24,3 +33,136 @@ func TestOperatorNamespace(t *testing.T) {
...
+	}
+	var pods *corev1.PodList
+	testTimeout := time.Second * 120
+	// Poll to see if pod is created
+	wait.PollImmediate(time.Second*1, testTimeout, func() (bool, error) {
+		pods, err = kubeClient.CoreV1().Pods(name).List(metav1.ListOptions{LabelSelector: "app=pdbtest"})
+		if err != nil {
+			return false, err
+		}
+		if len(pods.Items) > 0 && pods.Items[0].Status.Phase == corev1.PodRunning {
```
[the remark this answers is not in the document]
soltysh: This still holds.

test/e2e/operator_test.go
```diff
@@ -61,3 +164,46 @@ func TestKCMRecovery(t *testing.T) {
...
+						"sh",
+						"-c",
+						"trap exit TERM; while true; do sleep 5; done",
+					},
+				},
+			},
+		},
+	}
+	_, err := client.CoreV1().Pods(name).Create(pod)
+	if err != nil {
```
soltysh: Just `return err`

Makefile
sallyom: I prefer to see what's going on rather than a hang, but I removed this change : )

test/e2e/operator_test.go
sallyom: done

PR #320, opened 2019-12-06
Specify "bindNetwork: tcp" so controller manager will work on IPv6 single-stack

danwinship:
If you say `net.Listen("tcp", "0.0.0.0:10357")`, golang will actually create a dual-stack IPv4/IPv6 listening socket. But for historical reasons (having to do with bad IPv6 compat in older kube I think), we default to passing `"tcp4"` (the default value of `configv1.ServingInfo.BindNetwork`) instead of `"tcp"`, forcing IPv4-only.

There is no reason these days for servers on the pod network to enforce IPv4-only, so this overrides the default for kube-controller-manager.

This will need to be backported to 4.3.

(based on an earlier patch from @russellb )


russellb:
/lgtm

Thank you!

soltysh:
/cherry-pick release-4.3
@danwinship for 4.3 you'll need a BZ

PR #333, opened 2020-01-07
Bug 1788827: Condition to suppress PodDisruptionBudgetAtLimit alert when no disruption is allowed

### LINKED ISSUE #332: PodDisruptionBudgetAtLimit alert should not be raised when there no disruption allowed.
PodDisruptionBudgetAtLimit should not be raised when there is no disruption allowed: 
Example:
When there is only one pod and no disruption is allowed on it, this alert keeps on showing all the time. 

In this kind of situation, this alert should not be raised, as it can be a conscious decision to have the expected number of pods to be equal to the min number of pods, and not allow any disruption.

anmolsachan:
Fixes #332 

Signed-off-by: Anmol Sachan <anmol13694@gmail.com>

### DIFF
--- a/manifests/0000_90_kube-controller-manager-operator_05_alert-pdb.yaml
+++ b/manifests/0000_90_kube-controller-manager-operator_05_alert-pdb.yaml
@@ -13,7 +13,7 @@ spec:
           annotations:
             message: The pod disruption budget is preventing further disruption to pods because it is at the minimum allowed level.
           expr: |
-            kube_poddisruptionbudget_status_expected_pods == on(namespace, poddisruptionbudget, service) kube_poddisruptionbudget_status_desired_healthy
+            kube_poddisruptionbudget_status_expected_pods == on(namespace, poddisruptionbudget, service) kube_poddisruptionbudget_status_desired_healthy and on(namespace, poddisruptionbudget, service) kube_poddisruptionbudget_status_pod_disruptions_allowed >0
           for: 15m
           labels:
             severity: warning


smarterclayton:
/hold

What’s the change doing?  Please describe why this change fixes the issue.

smarterclayton:
The alert exists because this condition will block upgrades.  It’s possible we should allow the pdb to be modified in such a way that the alert is suppressed - however my next question is:

Why do you have a PDB with no disruptions allowed?

eparis:
As I recall, OCS has a PDB which does not allow any disruption. The also have a tripwire pod not covered by any PDB. If the tripwire pod is drained then the OCS operator makes a decision if it is safe to drain that node from an OCS PoV. If it is safe then OCS will change the PDB.  The decision on if a particular pod can be drained however is too complex to describe in a PDB (it involves global information)

rohantmp:
We have a special case in Rook/OCS where we detect drains using canaries and selectively unblock them if they won't cause disruption. We also block drains until the data is sufficiently replicated that it can take some disruption. It was the only way that Ceph could remain HA during a rolling upgrade.

Essentially:

- a drain is attempted and blocked by default because of the restrictive PDBs
- the drain is detected and allowed(by deleting the PDBs for that set of daemons) if the cluster is healthy enough
- the drain succeeds
- Ceph storage has a partial disruption that it can easily handle
- a second drain is attempted elsewhere and is blocked because of a restrictive pdb there
- we wait for data to be sufficiently replicated and delete the PDBs for that set of daemons
- and repeat

In this way, the upgrade succeeds and Ceph IO never suffers downtime caused by drains.
It's not pretty, but PDBs are otherwise insufficient to represent the constraints of stateful apps.

eparis:
Maybe we could put annotation on the PDB or something extra like that to say 'this PDB is managed by another global state controller.'     We added this alert because istio/service mesh did PDBs wrong and broke customer upgrades. The "solution" presented here would, in fact, make the alert inadequate for the istio mistake...

anmolsachan:
> /hold
> 
> What’s the change doing? Please describe why this change fixes the issue.

The change adds a condition, which allows only those pod alerts to trigger which have disruption allowed >0.
It takes care of the condition, where the user does not want any disruption to happen. Example,
For a pod which has 1 replica by default and it should always stay and not be disrupted. 
The condition just allows alerts to raise for those pods, where at least one disruption is allowed.
Similarly, this can happen for 2-2 combination or 3-3 combination, where disruption is not at all allowed and all the pods need to stay always.

soltysh:
If you're already twiddling with adding and removing PDBs, why not making them relevant by setting the minimal disruption?

anmolsachan:
> Maybe we could put annotation on the PDB or something extra like that to say 'this PDB is managed by another global state controller.' We added this alert because istio/service mesh did PDBs wrong and broke customer upgrades. The "solution" presented here would, in fact, make the alert inadequate for the istio mistake...

@eparis The condition only filters the alerts for the pods which cannot have any disruption. Would that fail the purpose of the actual alert?
In the condition when the deployed number of pods is equal to min number of pods, this alerts would always show up, but that is the deployment choice user has made.

rohantmp:
> why not making them relevant by setting the minimal disruption?

I don't understand what you mean by minimal disruption. The minimal disruption is none at all and that's what we do, and anymore would break the design.

Suppressing the alert for marked PDBs works.

eparis:
any user application can create a PDB with a disruption == 0 which was not being controlled by some other gloal state controller. Such a PDB (which was done by service mesh) would prevent upgrades. Such a PDB should be alerted.

This PR would ignore those types of PDBs. AKA a type which would definitely break upgrades.

The fact that OCS has a secondary controller and that it's PDB with disruption==0 is "special" should be marked as special. And the alert should ignore PDBs with such a marker. But ignoring "most" PDBs which are going to break upgrades is not acceptable.

eparis:
I was a second, EXPLICIT, mark on a PDB that says 'I know this has a high likelihood of causing problems and I don't want to alert about it anyway.'  Thus people who unknowingly/accidentally create such PDBs will get noticed and fixed.

anmolsachan:
@eparis I understand your explanation it is it valid. 
To solve this issue: For the alert to ignore PDB's which are marked as special, I suppose we will have to add a Special annotation on the PDB, which then will have to picked by the Prometheus exporter(kube-state-metrics). Then we modify the alert on the basis of that to ignore special cases. Is my understanding right on this? 
Also, this would require a change in the kube-state-metrics as well?

smarterclayton:
The annotation is probably the most correct workaround we can do right now, but requires us to change the ksm code.

I’m pretty sure ocs is abusing PDB due to a gap in PDB function.  The gap needs to get fixed independent of this PR.  The gap might be that we need a way to ensure finalization of a specific node (rebalancing) can complete independent of pod lifecycle.

anmolsachan:
Since this PR won't be a good fix, closing it.

PR #341, opened 2020-02-04
Move default config to its own path

marun:
/cc @sttts


marun:
/hold

Waiting for 4.5, this is not worth filing a bz for.

PR #342, opened 2020-02-04
Bug 1798048: bindata: remove revision suffix from containers

mfojtik:
We don't need revision suffix as the pod already has revision in label. The original intent was to make it clear which revision we have in artifact logs, but having this suffix makes it harder to debug or to get logs when debugging broken clusters.

Combined with kubernetes/kubernetes#87809 I think we will get into desirable state :-)

/cc @deads2k
/cc @smarterclayton


smarterclayton:
Yay

PR #351, opened 2020-02-25
Bug 1804239: Make csr-signer live-reloaded

tnozicka:
csr-signer is wired for both live-reload and revisioned, but revisioned refs aren't used. Live reloading went in https://github.com/openshift/origin/pull/24577 . If we live reload we don't want to trigger revision rollout.

/cc @p0lyn0mial @soltysh @deads2k


tnozicka:
failed to acquire lease: status 503 Service Unavailable, status code 503
/retest

PR #360, opened 2020-03-06
Add clusterrole for namespace security allocatoion controller

stlaz:
Takes the roles from the openshift-apiserver repo and moves them here, these are required for the controller to function so this repository feels like the correct spot where the clusterrole and its binding should reside.

/cc @sallyom @soltysh @sttts


sttts:
/approve
lgtm

stlaz:
/test e2e-aws
getting 400s

PR #363, opened 2020-03-11
Bug 1812583: Normalize CPU requests on masters

smarterclayton:
The kcm uses approximately 5% of master CPU in a reasonable medium
sized workload. The kcm can be bursty in certain scenarios.

Given a 1 core per master baseline (since CPU is compressible and
shared), assign the kcm roughly 10% of that core on each master
(1/3 of the kube-apiserver usage).


smarterclayton:
4.4 bug is 1812709

PR #379, opened 2020-03-18
reduce cert rotation scale for csr-signer

vareti:
with this change, we will check the possibility to test start/stop procedure within minutes instead of waiting for hours. Right now, scaling factor of 8 limits this as `csr-signer` would rotate at a different interval than other control plane certs.


vareti:
/close

this PR is not actively worked on anymore.

PR #382, opened 2020-03-30
Bug 1818420: delay only when csr-signer is from the future

soltysh:
/assign @deads2k


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -496,7 +496,7 @@ func ManageCSRSigner(ctx context.Context, lister corev1listers.SecretLister, cli
 	// make sure we wait five minutes to propagate the change to other components, like kas for trust
 	useAfter := signingCertKeyPair.Config.Certs[0].NotBefore.Add(5 * time.Minute)
```
deads2k: I'm trying to replay in my mind what I was thinking.  I think the logic is like this:

```go
// compute a useAfter date
// get the current secret
switch {
case IsNotFound(signer):
    // apply the secret

case signer-is-expired:
    // apply the secret

case now is after useAfter:
    // apply the secret

default:
    // wait a little while longer until afte the use after
}


```


I think your code change is correct, but do you think this would be easier as switch/case?

soltysh: @deads2k have a look at it now and lemme know is this is more to your liking

PR #383, opened 2020-03-31
Bug 1819256: Fix race for localhost-recovery-client-token snapshotting

tnozicka:
When `localhost-recovery-client-token` secret got created there is a race between kube-controller-manager to inject a token into it and revision controller that snapshots it. If it snapshots it before the token is injected we deploy without a token. 

Requires https://github.com/openshift/library-go/pull/764

/cc @deads2k @soltysh @sttts

### DIFF
--- a/vendor/github.com/openshift/library-go/pkg/operator/resource/resourceapply/core.go
+++ b/vendor/github.com/openshift/library-go/pkg/operator/resource/resourceapply/core.go
@@ -348,6 +348,24 @@ func SyncSecret(client coreclientv1.SecretsGetter, recorder events.Recorder, sou
 	case err != nil:
 		return nil, false, err
 	default:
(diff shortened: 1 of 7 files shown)


tnozicka:
I'll replace this with smarter snapshotting

vendor/github.com/openshift/library-go/pkg/operator/resource/resourceapply/core.go
```diff
@@ -348,6 +348,24 @@ func SyncSecret(client coreclientv1.SecretsGetter, recorder events.Recorder, sou
 	case err != nil:
 		return nil, false, err
 	default:
+		if source.Type == corev1.SecretTypeServiceAccountToken {
```
tnozicka, line 351: here is the fix

PR #385, opened 2020-04-03
Bug 1820610: Refresh csr-signer in recovery flow

tnozicka:
We are already syncing the by products but the main cert rotation loop got left somewhere. 

/cc @deads2k @soltysh

### DIFF
--- a/pkg/cmd/recoverycontroller/cmd.go
+++ b/pkg/cmd/recoverycontroller/cmd.go
@@ -9,7 +9,11 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/pkg/version"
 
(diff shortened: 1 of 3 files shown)


deads2k:
there is logic that plumbs the CSR signer to the right spot for the cert syncer . it's not the same location because we have to strip our the extra cert.  You got that too?

/assign @soltysh

deads2k:
This bit here: https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L468-L514 stands out.

deads2k:
oh, and this one too: https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L453-L466

PR #388, opened 2020-04-06
[release-4.4] Bug 1813512: Fix race for localhost-recovery-client-token snapshotting

tnozicka:
backporting https://github.com/openshift/cluster-kube-controller-manager-operator/pull/383

/cc @soltysh


sdodson:
This is preventing us from getting solid signal on upgrades so I've overridden the bugzilla-valid flag ahead of having QE test the 4.5 bug.

PR #389, opened 2020-04-07
Bug 1821680: Sync csr-controller-ca to openshift-config-managed namespace

tnozicka:
When recovering from expired certificates, the refreshed `csr-controller-ca` configmap needs to be propagated to openshift-config-managed namespace to be consumed by CKAO to trust the new client certs.

/cc @deads2k @soltysh

### DIFF
--- a/pkg/cmd/recoverycontroller/cmd.go
+++ b/pkg/cmd/recoverycontroller/cmd.go
@@ -10,12 +10,12 @@ import (
 	"k8s.io/client-go/pkg/version"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
-	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/certrotationcontroller"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
 	"github.com/openshift/library-go/pkg/operator/certrotation"
 	"github.com/openshift/library-go/pkg/operator/genericoperatorclient"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 
+	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/certrotationcontroller"
 	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/operatorclient"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/cmd/recoverycontroller/cmd.go
```diff
@@ -116,6 +117,27 @@ func (o *Options) Run(ctx context.Context) error {
 		return err
 	}
 
+	resourceSyncController := resourcesynccontroller.NewResourceSyncController(
```
deads2k, line 120: find a way to restructure so that we can re-use the same controller for operator and here.

pkg/cmd/recoverycontroller/cmd.go
```diff
@@ -116,6 +117,27 @@ func (o *Options) Run(ctx context.Context) error {
...
+		operatorClient,
+		kubeInformersForNamespaces,
+		v1helpers.CachedSecretGetter(kubeClient.CoreV1(), kubeInformersForNamespaces),
+		v1helpers.CachedConfigMapGetter(kubeClient.CoreV1(), kubeInformersForNamespaces),
+		o.controllerContext.EventRecorder,
+	)
+	err = resourceSyncController.SyncConfigMap(
+		resourcesynccontroller.ResourceLocation{
+			Namespace: operatorclient.GlobalMachineSpecifiedConfigNamespace,
+			Name:      "csr-controller-ca",
```
deads2k, line 130: where are you calling ManageCSRCABundle ?

pkg/operator/resourcesynccontroller/resourcesynccontroller.go
```diff
@@ -10,6 +10,13 @@ import (
 	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/operatorclient"
 )
 
+func AddSyncCSRControllerCA(resourceSyncController *resourcesynccontroller.ResourceSyncController) error {
```
soltysh, line 13: Nit, I'd call this `SyncCSRControllerCA` you're not adding anything.

tnozicka: It is adding new rule to `resourceSyncController` object underneath for syncing the ConfigMap :)
Given this has to be a non-member method I wanted to avoid calling it `SyncCSRControllerCA` which would imply this actually does the sync.

PR #394, opened 2020-04-11
Bug 1820687: allow patch for updating namespace

soltysh:
/assign @deads2k

### DIFF
--- a/pkg/operator/v411_00_assets/bindata.go
+++ b/pkg/operator/v411_00_assets/bindata.go
@@ -689,27 +689,15 @@ metadata:
     revision: "REVISION"
 spec:
   initContainers:
-  - name: wait-for-host-port
+  - name: wait-for-host-ports
     terminationMessagePolicy: FallbackToLogsOnError
     image: ${IMAGE}
     imagePullPolicy: IfNotPresent
-    command: ['/usr/bin/timeout', '30', "/bin/bash", "-c"]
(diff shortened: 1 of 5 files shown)


test/e2e/satokensigner_test.go
```diff
@@ -27,17 +27,20 @@ func TestSATokenSignerControllerSyncCerts(t *testing.T) {
 	ctx := context.Background()
 
 	// wait for the operator readiness
+	t.Logf("Waiting for true, false, false")
```
tnozicka, line 30: t.Log, until 1.14, is not logging until the test finishes but if it panics we don't have logs - I'd do klog.Infof for now (don't forget klog.Init

pkg/operator/v411_00_assets/bindata.go
```diff
@@ -688,27 +689,15 @@ metadata:
...
-        echo -n "."
-        sleep 1
-      done
-  - name: wait-for-cpc-host-port
+  - name: wait-for-host-ports
     terminationMessagePolicy: FallbackToLogsOnError
     image: ${IMAGE}
     imagePullPolicy: IfNotPresent
-    command: ['/usr/bin/timeout', '30', "/bin/bash", "-c"]
+    command: ['/usr/bin/timeout', '105', '/bin/bash', '-ec'] # a bit more than 60s for graceful termination + 35s for minimum-termination-duration, 5s extra cri-o's graceful termination period
```
tnozicka, line 696: graceful termination applies to apiserver, I think we should set 60 s for TIME_WAIT + 5 s for crio

PR #400, opened 2020-04-20
Bug 1825983: Sync new client cert-key on recovery

tnozicka:
We need new client keys to be able to start KCM before operators start to fix the mess.

On recovery flow, KASO regenerate certificates routine create new KCM client certs in openshift-config-managed namespace. We need to sync them to openshift-kube-controller-manager namespace so cert syncer can sync them to disk to be live-reloaded.

/cc @deads2k @soltysh

### DIFF
--- a/pkg/cmd/recoverycontroller/csrcontroller.go
+++ b/pkg/cmd/recoverycontroller/csrcontroller.go
@@ -88,6 +88,10 @@ func NewCSRController(
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


soltysh:
/hold
we need https://github.com/openshift/library-go/pull/775 fixed

tnozicka:
library-go fix merged in https://github.com/openshift/cluster-kube-controller-manager-operator/pull/402
/hold cancel

/retest
(PDB flake)

PR #401, opened 2020-04-20
[release-4.4] Bug 1825987: Sync new client cert-key on recovery

tnozicka:
backport of https://github.com/openshift/cluster-kube-controller-manager-operator/pull/400

/cc @deads2k @soltysh @mfojtik

### DIFF
--- a/pkg/cmd/recoverycontroller/csrcontroller.go
+++ b/pkg/cmd/recoverycontroller/csrcontroller.go
@@ -88,6 +88,10 @@ func NewCSRController(
 	if err != nil {
 		return nil, err
 	}
+	err = operatorresourcesync.AddSyncClientCertKeySecret(c.resourceSyncController)
+	if err != nil {
+		return nil, err
+	}
 
 	return c, nil
 }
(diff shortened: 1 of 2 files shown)


tnozicka:
/assign @mfojtik 
for cherry-pick label

tnozicka:
@sdodson this should go in asap, anything blocking labeling this?

sdodson:
@tnozicka Does https://github.com/openshift/cluster-kube-controller-manager-operator/pull/402 need to be backported? otherwise this looks fine. Upstream change merged Apr 22, was verified Apr 26.

tnozicka:
> @tnozicka Does #402 need to be backported?

@sdodson Doesn't look like they managed to break it for 4.4 (only master) https://github.com/openshift/cluster-ingress-operator/blob/3d9a5d3/pkg/operator/controller/names.go#L62

tnozicka:
The connection to https://github.com/openshift/cluster-kube-controller-manager-operator/pull/402 on the upstream PR was that is was blocking master merges at the time because the ingress rename PR broke it and that job wasn't gating other repos.

PR #405, opened 2020-04-27
Bug: 1805908: Don't use upstream recycler pod

bertinatto:
If no recycler pod template is specified, the controller manager will use a default one. However, we can't rely on that one because:

1) The pod is started in the `default` namespace
2) It uses the `busybox` image

Note: the `Recycle` reclaim policy was deprecated long ago in favour of dynamic provisioning [1], but if a user happens to use this feature, we'll hit the points above.

CC @openshift/storage 

[1] https://docs.openshift.com/container-platform/4.4/storage/understanding-persistent-storage.html#reclaiming_understanding-persistent-storage

### DIFF
--- a/bindata/bootkube/config/bootstrap-config-overrides.yaml
+++ b/bindata/bootkube/config/bootstrap-config-overrides.yaml
@@ -21,3 +21,7 @@ extendedArguments:
   service-cluster-ip-range: {{range .ServiceClusterIPRange}}
   - {{.}}{{end}}
   {{end}}
+  pv-recycler-pod-template-filepath-nfs: # bootstrap KCM doesn't need recycler templates
+  - ""
+  pv-recycler-pod-template-filepath-hostpath:
+  - ""
--- a/bindata/v4.1.0/config/defaultconfig.yaml
+++ b/bindata/v4.1.0/config/defaultconfig.yaml
@@ -15,6 +15,10 @@ extendedArguments:
   - "true"
   flex-volume-plugin-dir:
   - "/etc/kubernetes/kubelet-plugins/volume/exec" # created by machine-config-operator, owned by storage team/hekumar@redhat.com
+  pv-recycler-pod-template-filepath-nfs:
+  - "/etc/kubernetes/manifests/recycler-pod.yaml" # created by machine-config-operator, owned by storage team/fbertina@redhat.com
+  pv-recycler-pod-template-filepath-hostpath:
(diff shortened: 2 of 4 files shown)


bindata/bootkube/config/bootstrap-config-overrides.yaml
```diff
@@ -21,3 +21,7 @@ extendedArguments:
   service-cluster-ip-range: {{range .ServiceClusterIPRange}}
   - {{.}}{{end}}
   {{end}}
+  pv-recycler-pod-template-filepath-nfs:
+  - ""
+  pv-recycler-pod-template-filepath-hostpath:
+  - ""
```
bertinatto, line 27: The bootstrap KCM won't find the template file.

@smarterclayton should I override the config here or should I create the file (in the installer?)?

bertinatto:
/test e2e-aws-operator

The failing test seems unrelated to this PR. This BZ might indicate it's a flake: https://bugzilla.redhat.com/show_bug.cgi?id=1835992

bertinatto:
Temporarly increased test timeout to check if that's the reason for the failures.

bertinatto:
> Temporarly increased test timeout to check if that's the reason for the failures.

Apparently this helped. Now I removed the change (timeout increase) from this PR and if we hit the failure again we'll look into merging https://github.com/openshift/cluster-kube-controller-manager-operator/pull/425

bertinatto:
/assign @soltysh 

Maciej, could you take a look and let me know if this makes sense?

jsafrane:
We don't want to use upstream recycler because it uses busybox and we we don't support busybox. This busybox recycler image can still be used during the cluster bootstrap phase though. Do we support storage provisioning / recycler or actually anything about Kubernetes API at this time? I hope we don't (wait for the installation to complete!)

bindata/v4.1.0/config/defaultconfig.yaml
```diff
@@ -15,6 +15,10 @@ extendedArguments:
   - "true"
   flex-volume-plugin-dir:
   - "/etc/kubernetes/kubelet-plugins/volume/exec" # created by machine-config-operator, owned by storage team/hekumar@redhat.com
+  pv-recycler-pod-template-filepath-nfs: # avoid using default template, even though recycler reclaim policy is deprecated
+  - "/etc/kubernetes/manifests/recycler-pod.yaml"
```
soltysh, line 19: Where this file comes from? I assume you guys are creating it, can you make a comment similar to the one above?

bertinatto: Sure, added comment to state it's created by MCO

PR #411, opened 2020-05-08
Bug 1813894: Enable the addition of service ca to token secrets via UnsupportedConfigOverride

marun:
Previously the operand always added the serving service ca to token secrets. 

https://github.com/openshift/origin/pull/24956 will disable this behavior by default but allow it to be enabled via an env var. 

This PR supports setting the env var if an `UnsupportedConfigOverride` named `enableDeprecatedAndRemovedServiceCAKeyUntilNextRelease_ThisMakesClusterImpossibleToUpgrade` is set to true. 

When the override is set to true, the operator will be marked as not upgradeable. This is a temporary measure to allow but discourage the behavior prior to its removal in 4.6.

/cc @sttts @deads2k @soltysh


marun:
/hold 

Waiting on https://github.com/openshift/library-go/pull/792 to merge

marun:
/retest

Flakes!

marun:
Rebased

PR #414, opened 2020-05-12
Bug 1806640: fix potential errors in Prometheus alerts

simonpasquier:
There is no need to use the 'on' keyword to compare
kube_poddisruptionbudget_status_expected_pods and
kube_poddisruptionbudget_status_desired_healthy because both series have
the same label set. In addition when there are 2 instances of
kube-state-metrics running, it triggers evaluation errors in Prometheus.

To ensure that the expression returns one item per (namespace,
poddisruptionbudget) tuple, it is wrapped by the 'max' aggregation
operator.

### DIFF
--- a/manifests/0000_90_kube-controller-manager-operator_05_alert-pdb.yaml
+++ b/manifests/0000_90_kube-controller-manager-operator_05_alert-pdb.yaml
@@ -13,7 +13,7 @@ spec:
           annotations:
             message: The pod disruption budget is preventing further disruption to pods because it is at the minimum allowed level.
           expr: |
-            kube_poddisruptionbudget_status_expected_pods == on(namespace, poddisruptionbudget, service) kube_poddisruptionbudget_status_desired_healthy
+            max by(namespace, poddisruptionbudget) (kube_poddisruptionbudget_status_expected_pods == kube_poddisruptionbudget_status_desired_healthy)
           for: 15m
           labels:
             severity: warning
@@ -21,7 +21,7 @@ spec:
           annotations:
             message: The pod disruption budget is below the minimum number allowed pods.
           expr: |
-            kube_poddisruptionbudget_status_expected_pods < on(namespace, poddisruptionbudget, service) kube_poddisruptionbudget_status_desired_healthy
+            max by (namespace, poddisruptionbudget) (kube_poddisruptionbudget_status_expected_pods < kube_poddisruptionbudget_status_desired_healthy)
           for: 15m
           labels:
             severity: critical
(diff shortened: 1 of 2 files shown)


manifests/0000_90_kube-controller-manager-operator_05_alert-pdb.yaml
```diff
@@ -11,17 +11,17 @@ spec:
       rules:
         - alert: PodDisruptionBudgetAtLimit
           annotations:
-            message: The pod disruption budget is preventing further disruption to pods because it is at the minimum allowed level.
+            message: The pod disruption budget {{ $labels.namespace }}/{{ $labels.poddisruptionbudget }} is preventing further disruption to pods because it is at the minimum allowed level.
```
soltysh, line 14: We don't support variables in our rendering pipeline, yet so I'd suggest dropping this. When there will be a need we'll add them. For now it'll be simpler w/o those and will clear the unit failure.

manifests/0000_90_kube-controller-manager-operator_05_alert-pdb.yaml
```diff
@@ -11,17 +11,17 @@ spec:
...
           expr: |
-            kube_poddisruptionbudget_status_expected_pods == on(namespace, poddisruptionbudget, service) kube_poddisruptionbudget_status_desired_healthy
+            max by(namespace, poddisruptionbudget) (kube_poddisruptionbudget_status_expected_pods == kube_poddisruptionbudget_status_desired_healthy)
```
soltysh, line 16: Since you're changing how this alert is reporting you'll need to update e2e in https://github.com/openshift/cluster-kube-controller-manager-operator/blob/57aab7b4c200f43847f66195d3777b638ddda65a/test/e2e/operator_test.go#L124 to also drop the `service="kube-state-metrics"` bit too.

manifests/0000_90_kube-controller-manager-operator_05_alert-pdb.yaml
simonpasquier: Thanks for the info, I wasn't aware that the manifests could be templatized. I'll rollback to the previous version.
The templating here was meant for the Prometheus server when it evaluates the alerting rule. See the Prometheus [documentation](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/#templating).

manifests/0000_90_kube-controller-manager-operator_05_alert-pdb.yaml
simonpasquier: :+1:

PR #420, opened 2020-06-11
wait for localhost kas to respond

deads2k:
just in case graceful termination made new connection logic more stringent.

### DIFF
--- a/bindata/v4.1.0/kube-controller-manager/pod.yaml
+++ b/bindata/v4.1.0/kube-controller-manager/pod.yaml
@@ -35,6 +35,20 @@ spec:
             echo "Copying system trust bundle"
             cp -f /etc/kubernetes/static-pod-certs/configmaps/trusted-ca-bundle/ca-bundle.crt /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
           fi
+
+          echo -n "Waiting kube-apiserver to respond."
+          tries=0
+          until curl --output /dev/null --silent -k https://localhost:6443/version; do
+            echo -n "."
+            sleep 1
+            (( tries += 1 ))
+            if [[ "${tries}" -gt 180 ]]; then
+              echo "timed out waiting for kube-apiserver to respond."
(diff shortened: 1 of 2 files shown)


bindata/v4.1.0/kube-controller-manager/pod.yaml
```diff
@@ -35,6 +35,20 @@ spec:
             echo "Copying system trust bundle"
             cp -f /etc/kubernetes/static-pod-certs/configmaps/trusted-ca-bundle/ca-bundle.crt /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
           fi
+
+          echo -n "Waiting kube-apiserver to respond."
+          tries=0
+          until $(curl -k https://localhost:6443/version); do
```
retroflexer, line 41: Please remove the `$`. The following command should work:

```
until curl --output /dev/null --silent -k https://localhost:6443/version; do
```

soltysh:
/hold
The same comment as in https://github.com/openshift/cluster-kube-scheduler-operator/pull/254#issuecomment-644839984 applies here, even though it might be pre-existing.

soltysh:
/hold cancel
let's get this as is, and we'll need a proper long-term solution but one that won't be backported and this needs backports.

bindata/v4.1.0/kube-controller-manager/pod.yaml
```diff
@@ -35,6 +35,20 @@ spec:
             echo "Copying system trust bundle"
             cp -f /etc/kubernetes/static-pod-certs/configmaps/trusted-ca-bundle/ca-bundle.crt /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
           fi
+
+          echo -n "Waiting kube-apiserver to respond."
+          tries=0
+          until curl --output /dev/null --silent -k https://localhost:6443/version; do
```
sttts, line 41: why version? We have `/readyz` for exactly this purpose.

PR #418, opened 2020-06-03
Use ss instead of lsof when waiting for ports

soltysh:
This is coming from this change to kas-o https://github.com/openshift/cluster-kube-apiserver-operator/pull/864 for details why `ss` over `lsof` see this comment https://github.com/openshift/cluster-kube-apiserver-operator/pull/864#issuecomment-632322430.

/assign @sttts

PR #421, opened 2020-06-23
Fix port check

tnozicka:
Container can be restarted while the pods stays. For that reason, we need to check the port availability in the same process as we listen, not in an init container which isn't re-run.

Adding startup probe to avoid liveness probe timeout and killing the process.

/cc @soltysh @sttts

### DIFF
--- a/bindata/v4.1.0/kube-controller-manager/pod.yaml
+++ b/bindata/v4.1.0/kube-controller-manager/pod.yaml
@@ -10,27 +10,16 @@ metadata:
     kube-controller-manager: "true"
     revision: "REVISION"
 spec:
-  initContainers:
-  - name: wait-for-host-ports
-    terminationMessagePolicy: FallbackToLogsOnError
-    image: ${IMAGE}
-    imagePullPolicy: IfNotPresent
-    command: ['/usr/bin/timeout', '65', '/bin/bash', '-ec'] # a bit more than 60s for TIME_WAIT, 5s extra cri-o's graceful termination period
-    args:
-    - |
-      echo -n "Waiting for port :10257 and :10357 to be released."
-      while [ -n "$(lsof -ni :10257)$(lsof -ni :10357)" ]; do
-        echo -n "."
-        sleep 1
-      done
   containers:
   - name: kube-controller-manager
     image: ${IMAGE}
     imagePullPolicy: IfNotPresent
     terminationMessagePolicy: FallbackToLogsOnError
-    command: ["/bin/bash", "-ec"]
+    command: ["/bin/bash", "-euxo", "pipefail", "-c"]
     args:
         - |
+          timeout 3m /bin/bash -exuo pipefail -c 'while [ -n "$(ss -Htanop \( sport = 10257 \))" ]; do sleep 1; done'
+
           if [ -f /etc/kubernetes/static-pod-certs/configmaps/trusted-ca-bundle/ca-bundle.crt ]; then
             echo "Copying system trust bundle"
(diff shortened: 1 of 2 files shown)


bindata/v4.1.0/kube-controller-manager/pod.yaml
```diff
@@ -10,27 +10,17 @@ metadata:
...
-        echo -n "."
-        sleep 1
-      done
   containers:
   - name: kube-controller-manager
     image: ${IMAGE}
     imagePullPolicy: IfNotPresent
     terminationMessagePolicy: FallbackToLogsOnError
-    command: ["/bin/bash", "-ec"]
+    command: ["/bin/bash", "-euxo", "pipefail", "-c"]
```
tnozicka, line 18: with -x I suppose the echo statement are not needed

bindata/v4.1.0/kube-controller-manager/pod.yaml
```diff
@@ -10,27 +10,17 @@ metadata:
...
     args:
         - |
+          echo -n "Waiting for port :10257 to be released."
+          timeout 3m /bin/bash -euo pipefail -c 'while [ -n "$(lsof -ni :10257)" ]; do echo -n "."; sleep 1; done'
```
sttts, line 22: please don't diverge from https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/v4.1.0/kube-apiserver/pod.yaml. This lsof does not work.

bindata/v4.1.0/kube-controller-manager/pod.yaml
sttts: this diverges from kube-apiserver operator. I don't like this.

bindata/v4.1.0/kube-controller-manager/pod.yaml
tnozicka: I have not noticed you changed it since then. What didn't work about the lsof? I have tried it locally but sounds like there is corner case?

bindata/v4.1.0/kube-controller-manager/pod.yaml
tnozicka: Might worth to fix it there too - `-u` and `-o pipefail` are good practice to prevent mistakes and typos.

`-x` is a matter of opinion - less pretty output but 10x easier to reason about when debugging

bindata/v4.1.0/kube-controller-manager/pod.yaml
soltysh: Yeah, I had https://github.com/openshift/cluster-kube-controller-manager-operator/pull/418/ changing it but it didn't merge. @tnozicka can you pick that up here.

bindata/v4.1.0/kube-controller-manager/pod.yaml
soltysh: I'm :+1: for `-x`, it can also be added to kas-o, as well.

bindata/v4.1.0/kube-controller-manager/pod.yaml
```diff
@@ -87,6 +83,9 @@ spec:
     command: ["/bin/bash", "-ec"]
```
soltysh, line 83: Don't you want to change this one into `["/bin/bash", "-euxo", "pipefail", "-c"]` as well?

bindata/v4.1.0/kube-controller-manager/pod.yaml
tnozicka: converted to `ss`

bindata/v4.1.0/kube-controller-manager/pod.yaml
tnozicka: done

PR #425, opened 2020-06-30
Bug 1835992: Raise operator e2e test timeout

tnozicka:
`TestSATokenSignerControllerSyncCerts` is seen timing out after 10 minutes in https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-kube-controller-manager-operator/405/pull-ci-openshift-cluster-kube-controller-manager-operator-master-e2e-aws-operator/1277627343967358976/artifacts/junit_operator.xml

Let's see if it's just slower infra or the test is flaky.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ $(call build-image,ocp-cluster-kube-controller-manager-operator,$(IMAGE_REGISTRY
 $(call add-bindata,v4.1.0,./bindata/v4.1.0/...,bindata,v411_00_assets,pkg/operator/v411_00_assets/bindata.go)
 
 test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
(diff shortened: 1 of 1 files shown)


tnozicka:
CI seems to have recovered, but using 15 minutes instead of default 10 minutes wont hurt and will give us a place to adjust in the future.

/assign @soltysh

tnozicka:
actually, after taking a closer look, I've decided we should rather raise the timeout a lot to easily distinguish cases where infra is just slow to rollout the operator (for disruptive tests) and when it is actually stuck

bertinatto:
/retitle Bug 1835992: Raise operator e2e test timeout

Linking the BZ, feel free to remove it if it's not the case.

PR #428, opened 2020-07-02
Cleanup e2e tests and fix flakes

tnozicka:
Not limited to, but fixes SA issue from https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-kube-controller-manager-operator/426/pull-ci-openshift-cluster-kube-controller-manager-operator-release-4.4-e2e-aws-operator/1278566968177201152
That job was looking kind of bad in stats.
TestPodDisruptionBudgetAtLimitAlert [14d] Across 160474 runs and 4139 jobs (28.44% failed)
TestSATokenSignerControllerSyncCerts [14d] Across 106070 runs and 3787 jobs (29.61% failed)

/cc @soltysh @sallyom


test/e2e/operator_test.go
```diff
@@ -6,34 +6,29 @@ import (
...
-	policyclientv1beta1 "k8s.io/client-go/kubernetes/typed/policy/v1beta1"
-
-	routeclient "github.com/openshift/client-go/route/clientset/versioned"
-	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/operatorclient"
-	test "github.com/openshift/cluster-kube-controller-manager-operator/test/library"
-	"github.com/openshift/library-go/test/library/metrics"
-	"github.com/prometheus/common/model"
+	"k8s.io/client-go/tools/cache"
+	watchtools "k8s.io/client-go/tools/watch"
+	"k8s.io/klog"
```
tnozicka, line 27: TODO: raise the default loglevel in tests

test/library/helpers.go
```diff
@@ -0,0 +1,124 @@
...
+func RouteClient() *routeclient.Clientset {
+	routeClient, err := routeclient.NewForConfig(GetClientConfig())
+	if err != nil {
+		panic(err)
+	}
+
+	return routeClient
+}
+
+func ConfigClient() *configclient.Clientset {
```
soltysh, line 61: I'd prefer we don't introduce these above, they are simple enough not be helpers.

tnozicka: For me, the advantage is they avoid the error handling which differs when run as test only. It could also be done with `runtime.Must` pattern but I made it this way to follow the pattern we have in regular e2e's. I don't feel strong though - It's simplicity + consistency vs. deps which don't matter that much in test packages.

PR #431, opened 2020-07-21
use internal LB to avoid outages during kube-apiserver rollout

deads2k:
This makes us vulnerable to internal LB config that we do not own, but kubelets already break

Doing it like this leverages our existing config observation instead of creating a dynamic kubeconfig handling.

/assign @soltysh @sanchezl @tnozicka

### DIFF
--- a/pkg/operator/configobservation/masterurl/observe_masterurl_test.go
+++ b/pkg/operator/configobservation/masterurl/observe_masterurl_test.go
@@ -1,0 +1,118 @@
+package masterurl
+
+import (
+	"reflect"
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/cache"
+
+	configv1 "github.com/openshift/api/config/v1"
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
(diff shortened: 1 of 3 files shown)


deads2k:
I think this one is enough to fix the endpoints problem. We should move this to library-go if we decide on it for kube-scheduler too

pkg/operator/configobservation/masterurl/observe_masterurl_test.go
```diff
@@ -0,0 +1,118 @@
...
+			}
+			listers := configobservation.Listers{
+				InfrastructureLister_: configlistersv1.NewInfrastructureLister(indexer),
+			}
+			result, errs := ObserveMasterURL(listers, events.NewInMemoryRecorder("infraid"), test.input)
+			if len(errs) > 0 {
+				t.Fatal(errs)
+			} else {
+				if !reflect.DeepEqual(test.expected, result) {
+					t.Errorf("\n===== observed config expected:\n%v\n===== observed config actual:\n%v", toYAML(test.expected), toYAML(result))
```
tnozicka, line 105: cmp.Diff shows those nicely

tnozicka:
How does this work when the client kubeconfigs still point to localhost?

PR #437, opened 2020-08-25
WIP: Bug 1869962: Don't use upstream recycler pod

bertinatto:


sdodson:
/hold
We need to make sure that this includes fixes that check for the presence of the template because we cannot be assured that all versions upgrading into this code have the template deployed, ie: 4.3 to 4.4 upgrades would not unless they were upgrading from a version with the template.

jwforres:
This change should not merge without additional discussion about the known upgrade impacts. There is a discussion about figuring out an alternative fix that tolerates non-existence of the template to deal with upgrades from versions without the template.

PR #438, opened 2020-08-26
Bug 1868615: Revert "Don't use upstream recycler pod"

bertinatto:
This reverts commit 772810dce5557ea9521cc273d2d6c132a1d3d5a5 to prevent a CrashLoopBackOff if the recycler template isn't found (which can happen if this operator is upgraded before MCO).


sdodson:
A followup more complete fix will be filed ASAP but we'll do a speedy revert for now, marking cherry-pick-approved.

PR #440, opened 2020-09-01
Bug 1874532: using load balancer, there is no reason to wait on localhost

deads2k:
This ensures that the kcm and kas can rollout independently and successfully.


soltysh (approved):
/lgtm

PR #441, opened 2020-09-02
Bug 1867494: Explicitly use internal LB for KCM and CPC

tnozicka:
We should explicitly state which URL are we using instead of the hack with old master config override.
Also this wasn't wired for CPC and validating webhooks before.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1867494 and half of https://bugzilla.redhat.com/show_bug.cgi?id=1861917

/cc @soltysh @deads2k


soltysh (approved):
/lgtm
/hold
for @deads2k final sign off

PR #443, opened 2020-09-02
bug 1861917: add render --cluster-policy-controller-image to add that controller to  bootstrap yaml

deads2k:
This part 1 one of the three part series to render the cluster-policy-controller in bootstrap pod


soltysh (approved):
/lgtm

PR #448, opened 2020-09-14
bug 1874056: add debugging for machine-api, mco, and node teams

deads2k:
A demonstration of the code we're talking about.


deads2k:
it's a shame that we cannot get https://github.com/openshift/machine-config-operator/pull/2081 and https://github.com/openshift/machine-api-operator/pull/700 merged, but we can add the data while they quibble.

soltysh (approved):
/lgtm

PR #449, opened 2020-09-15
Bug 1861917: Fix bootstrap cpc

tnozicka:
CPC config was in a wrong location causing cpc on botstrap to crashloop.

Ness to be followed up with https://github.com/openshift/installer/pull/4178 and then enforcing the flag

/cc @soltysh


soltysh (approved):
/lgtm

soltysh (approved):
/lgtm

PR #450, opened 2020-09-16
[OCPCLOUD-1070] Set --cloud-provider=external for supported platforms

Danil-Grigorev:
Includes revendoring after https://github.com/openshift/library-go/pull/895 and https://github.com/openshift/library-go/pull/1077 changes are merged. Requires for KCM to apply `--cloud-provider=external` flag on supported platforms.


Danil-Grigorev:
/open

This is still very much required change.

go.mod
```diff
@@ -6,24 +6,25 @@ require (
...
+	k8s.io/api v0.20.0
+	k8s.io/apimachinery v0.20.0
+	k8s.io/apiserver v0.20.0
+	k8s.io/client-go v0.20.0
+	k8s.io/component-base v0.20.0
+	k8s.io/klog/v2 v2.4.0
 )
 
 replace (
+	github.com/openshift/library-go => github.com/Danil-Grigorev/library-go v0.0.0-20210217125431-c0ce7f525449
```
elmiko, line 27: this doesn't seem proper, shouldn't we have this pointing to the openshift/library-go after some patch has merged?

elmiko (commented):
confused about the replace in the go.mod

go.mod
Fedosin: we will bump the library here when https://github.com/openshift/library-go/pull/895 is merged

JoelSpeed:
IIUC, this revendor works because KCMO uses library go utilities (cloudprovider observer) already, so we just need to make sure it's using the latest version of that?

Danil-Grigorev:
> IIUC, this revendor works because KCMO uses library go utilities (cloudprovider observer) already, so we just need to make sure it's using the latest version of that?

Yes, that's correct @JoelSpeed

soltysh (approved):
/lgtm
/approve

PR #453, opened 2020-09-21
Bug 1861917: Fix bootstrap cpc config file location, certs and RBAC

tnozicka:
/cc @soltysh


tnozicka:
it still wait on the OAS
```
E0921 13:28:27.524881       1 namespace_scc_allocation_controller.go:258] the server could not find the requested resource (post rangeallocations.security.openshift.io)
```
but that's expected

soltysh (approved):
/lgtm

PR #454, opened 2020-09-22
Bug 1881246:  squash alerts into a single file

soltysh:
/assign @wking


manifests/0000_90_kube-controller-manager-operator_05_alert-pdb.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
     exclude.release.openshift.io/internal-openshift-hosted: "true"
 spec:
   groups:
-    - name: cluster-version
+    - name: kube-controller-manager
```
wking, line 10: Changing `spec.groups[].name` like this makes sense to me.  But the bug was about the overlap in `metadata.name`.  Can we combine the two PrometheusRule files in a single file with a single group with multiple rules?

soltysh: Gotcha.

soltysh:
@wking have a look at the current state, if that's closer to your preference?

wking (approved):
/lgtm

PR #455, opened 2020-09-22
Bug 1881534: Use internal load balancer instead of localhost 4.5

tnozicka:
/cc @soltysh @deads2k 

Backport of #441 (since #431 was reverted)


soltysh (approved):
/lgtm
/retest

markmc:
(patch manager) The PR linked from the 4.6 bug is #431 which is confusing. I see now that #431 was reverted and a new fix was merged in #441. I would have appreciated that explained clearly in the PR description. Added now.

markmc:
/retitle Bug 1881534: Use internal load balancer instead of localhost 4.5

PR #462, opened 2020-10-07
Bug 1886022: add new apigroup to rendered clusterrole copy

deads2k:
the rendered peer to https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/bindata/v4.1.0/kube-controller-manager/namespace-security-allocation-controller-clusterrole.yaml#L11


soltysh (approved):
/lgtm

PR #465, opened 2020-10-11
Support KubeVirt platform

ravidbro:
The purpose is to add support for KubeVirt platform which is supported with the new versions of library-go
* Bump github.com/openshift/library-go
* change network.GetServiceCIDR to network.GetServiceCIDRs, a change that was done in the new version of library-go

Signed-off-by: Ravid Brown <ravid@redhat.com>


ravidbro:
Closing the PR since 'library-go' was already bumped and included the change.

PR #469, opened 2020-10-19
Bug 1889308: Set dnsPolicy ClusterFirstWithHostNet to match hostNetwork

soltysh:
Based on https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy we should always set `dnsPolicy: ClusterFirstWithHostNet` when we have `hostNetwork: true`. 

/assign @tnozicka


damemi (approved):
/lgtm

PR #470, opened 2020-10-19
update for library-go API change (GetServiceCIDRs)

danwinship:
updates for openshift/library-go#925

This is utterly irrelevant, in that kcm's `--service-cluster-ip-range` argument currently only has any effect if (a) you're using `--allocate-node-cidrs` _and_ (b) the cluster CIDR and service CIDR overlap with each other (which I'm pretty sure openshift-install forbids). But maybe some day it will be used for something else, and we'll be ready for it!


danwinship:
this was pulled into #472

PR #471, opened 2020-10-20
Bug 1841119: Apply extendedArguments directly to kcm

soltysh:
The o/k changes are in https://github.com/openshift/kubernetes/pull/415

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -459,6 +459,18 @@ func managePod(ctx context.Context, configMapsGetter corev1client.ConfigMapsGett
 		containerArgsWithLoglevel[0] += " --tls-cert-file=/etc/kubernetes/static-pod-resources/secrets/serving-cert/tls.crt"
 		containerArgsWithLoglevel[0] += " --tls-private-key-file=/etc/kubernetes/static-pod-resources/secrets/serving-cert/tls.key"
 	}
+
+	kubeControllerManagerConfigMap, err := configMapsGetter.ConfigMaps(required.Namespace).Get(ctx, "config", metav1.GetOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, false, err
+	}
+	if kubeControllerManagerConfigMap != nil {
+		var kubeControllerManagerConfig map[string]interface{}
+		if err := yaml.Unmarshal([]byte(kubeControllerManagerConfigMap.Data["config.yaml"]), &kubeControllerManagerConfig); err != nil {
+			return nil, false, fmt.Errorf("failed to unmarshal the kube-controller-manager config: %v", err)
+		}
+		containerArgsWithLoglevel[0] += " " + readKubeControllerManagerArgs(kubeControllerManagerConfig)
+	}
 	containerArgsWithLoglevel[0] = strings.TrimSpace(containerArgsWithLoglevel[0])
 
 	var observedConfig map[string]interface{}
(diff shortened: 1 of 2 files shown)


soltysh:
/assign @deads2k @tnozicka 
for review

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -496,6 +506,23 @@ func managePod(ctx context.Context, configMapsGetter corev1client.ConfigMapsGett
...
+	}
+	args := []string{}
+	for key, value := range extendedArguments.(map[string]interface{}) {
+		for _, arrayValue := range value.([]interface{}) {
+			args = append(args, fmt.Sprintf("--%s=%s", key, arrayValue.(string)))
+		}
+	}
+	// make sure to sort the arguments, otherwise we might get mismatch
+	// when comparing revisions leading to new ones being created, unnecessarily
+	sort.Strings(args)
```
tnozicka, line 522: Have you checked the yaml parser version? I thought there was a bug that changed the order but with the correct version you'd not need to sort

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -496,6 +506,23 @@ func managePod(ctx context.Context, configMapsGetter corev1client.ConfigMapsGett
 	return resourceapply.ApplyConfigMap(configMapsGetter, recorder, configMap)
```
tnozicka, line 506: Please add a unit to make sure the args order and the configmap content in general is deterministic

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
soltysh: I don't think the yaml parser has anything to do with it. This sort operation is not that expensive and gives us necessary confidence.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
soltysh: Will do.

soltysh:
@tnozicka tests added ptal

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
tnozicka: For some programs order of arguments matter that's why I'd prefer to avoid the sort and rather figure out why the order changed while decoding. If I change arg order on the resource I'd expect it to rollout.

tnozicka: https://godoc.org/gopkg.in/yaml.v2#MapSlice

But since we control the args I can live with the sort as well.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -459,6 +459,16 @@ func managePod(ctx context.Context, configMapsGetter corev1client.ConfigMapsGett
 		containerArgsWithLoglevel[0] += " --tls-cert-file=/etc/kubernetes/static-pod-resources/secrets/serving-cert/tls.crt"
 		containerArgsWithLoglevel[0] += " --tls-private-key-file=/etc/kubernetes/static-pod-resources/secrets/serving-cert/tls.key"
 	}
+
+	kubeControllerManagerConfigMap, err := configMapsGetter.ConfigMaps(required.Namespace).Get(ctx, "config", metav1.GetOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, false, err
+	}
+	var kubeControllerManagerConfig map[string]interface{}
+	if err := yaml.Unmarshal([]byte(kubeControllerManagerConfigMap.Data["config.yaml"]), &kubeControllerManagerConfig); err != nil {
```
tnozicka, line 468: if it isn't found, `kubeControllerManagerConfigMap` could be nil

soltysh: since `Data` is `map[string]string` you can't have nil, at most it'll be an empty string. So that should work just fine.

tnozicka: the map itself `if err != nil && !apierrors.IsNotFound(err) `

soltysh: IC, fixed.

PR #478, opened 2020-11-03
picks up setting `TCP_USER_TIMEOUT` socket option

p0lyn0mial:
this operator uses `ControlexContext` that gets the `rest.Config` for communication with Kube API from `GetKubeConfigOrInClusterConfig` which in turn sets `TCP_USER_TIMEOUT` socket option.

please check https://github.com/openshift/library-go/pull/926 to find out why we want to set this option


soltysh (approved):
/lgtm
/approve
/retest

PR #482, opened 2020-11-18
bindata: move recycler manifest

sjenning:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1896226

This is the second step in resolving a bug introduced by https://github.com/openshift/machine-config-operator/pull/1687 and openshift/cluster-kube-controller-manager-operator#405

/etc/kubernetes/manifests is the kubelet's static pod manifests directory. The kubelet will try to start pods defined in this location as static pods. The recycler pod is not a static pod and the kubelet continuously fails to start it.

The solution is to move the recycler pod manifests to a different directory.

The fix is a 3-step process (as I see it):
- Change location of the recycler pod in the MCO template (https://github.com/openshift/machine-config-operator/pull/2238)
- Change KCM to use the new location (this PR)
- Delete (or empty out) the file in the old location

@runcom @sttts @rphillips @haircommander


sjenning:
/hold
until https://github.com/openshift/machine-config-operator/pull/2238 merges

sjenning:
/hold
on upgrade issue
/test e2e-aws

PR #494, opened 2021-01-12
Bug 1915500: support custom CA bundle for AWS C2S

staebler:
Communication with the AWS API in the C2S Secret region requires a custom CA bundle. This bundle is stored in the "ca-bundle.pem" entry in the cloud-config. The AWS_CA_BUNDLE environment variable informs the AWS SDK to use a custom CA bundle and to specify the location of the bundle.


staebler:
@deads2k This seems really ugly to me to inject cloud-specific logic into the pod yaml. Is there a better way to do this?

PR #499, opened 2021-01-25
Extend OWNERS

ingvagabund:
Extend list of approvers with @ingvagabund and @damemi


soltysh (approved):
/lgtm
/approve
This is updating OWNERS so I'm adding valid-bug label

PR #500, opened 2021-01-27
Bug 1920159: Adjust CPU request for controller manager more precisely

smarterclayton:
The actual use of the KCM in large production environments is
slightly lower than the request, and is fairly constant. In order
to ensure more headroom is available on small node sizes, tweak
downwards closer to the observed 40m real usage while still leaving
some headroom.

/hold

For testing with the other related changes

### DIFF
--- a/bindata/v4.1.0/kube-controller-manager/pod.yaml
+++ b/bindata/v4.1.0/kube-controller-manager/pod.yaml
@@ -39,7 +39,7 @@ spec:
     resources:
       requests:
         memory: 200Mi
-        cpu: 80m
+        cpu: 60m
     ports:
       - containerPort: 10257
     volumeMounts:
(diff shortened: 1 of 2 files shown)


smarterclayton:
/retest

Not related to the change

/hold cancel

Combined tests looked ok.  Team should review and see the comments in the bug.

soltysh (approved):
/lgtm

smarterclayton:
This repo's unit test jobs look broken.  I think this is a job that needs to use $ARTIFACT_DIR as steve noted in his aos-devel email.

smarterclayton:
Testing that theory in https://github.com/openshift/release/pull/15420

smarterclayton:
/retest

now that it is fixed

PR #506, opened 2021-02-18
bug 1844989: Allow to set --tls-cipher-suites and --tls-min-version of KCM instances through observed config

ingvagabund:
By allowing to set the ciphers, KCM does not have to use the default ciphers some of which may contain security vulnerabilities.

Blocked by https://github.com/openshift/cluster-policy-controller/pull/56

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -492,12 +492,32 @@ func managePod(ctx context.Context, configMapsGetter corev1client.ConfigMapsGett
 			containerArgsWithLoglevel[0] += " " + strings.Join(extendedArguments, " ")
 		}
 	}
-	containerArgsWithLoglevel[0] = strings.TrimSpace(containerArgsWithLoglevel[0])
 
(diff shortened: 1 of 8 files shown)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -492,12 +492,32 @@ func managePod(ctx context.Context, configMapsGetter corev1client.ConfigMapsGett
...
 		}
 	}
-	containerArgsWithLoglevel[0] = strings.TrimSpace(containerArgsWithLoglevel[0])
 
 	var observedConfig map[string]interface{}
 	if err := yaml.Unmarshal(operatorSpec.ObservedConfig.Raw, &observedConfig); err != nil {
 		return nil, false, fmt.Errorf("failed to unmarshal the observedConfig: %v", err)
 	}
+
+	cipherSuites, _, err := unstructured.NestedStringSlice(observedConfig, "servingInfo", "cipherSuites")
```
ingvagabund, line 501: @soltysh can we assume `.servingInfo` bits are always going to be set?

soltysh (approved):
/lgtm

PR #509, opened 2021-03-11
defines test-e2e-preferred-host job for running smoke tests to check if KCM talks to Kube over a preferred host

p0lyn0mial:


p0lyn0mial:
ref: https://github.com/openshift/kubernetes/pull/603

PR #517, opened 2021-04-09
Updates openshift/api

huffmanca:
The CSI Migration tech preview feature is included in `openshift/api`, and I'm bumping the version here so that the FeatureGate can be propagated correctly.

Goal of this PR is to apply `--feature-gate=CSIMigrationAWS=true,CSIMigrationOpenStack=true` to kubelet when `TechPreviewNoUpgrade featureSet` is enabled, which was introduced in openshift/api#880 .


huffmanca:
Closing this as https://github.com/openshift/cluster-kube-controller-manager-operator/pull/518/ was submitted.

PR #525, opened 2021-05-19
[OCPCLOUD-1109] Add ObserveCloudVolumePlugin config observer for KCM config

JoelSpeed:
This adds a new config observer to the KCM configuration observer.

This observer will set the `external-cloud-volume-plugin` option as appropriate when the cluster is running with an external cloud provider implementation.

This is required to allow KCM to attach and detach volumes that have not yet been migrated by the CSI migration process, while the rest of the cloud provider loops are no longer running.

Additionally, this updates the library-go dependency which in turn updates the cloud provider observer so that it also sets the `cloud-provider` configuration appropriately when the external cloud provider is running.

(This is currently in draft until https://github.com/openshift/library-go/pull/895 merges, at which point I will point the library go branch back to the openshift fork. My fork should match the tip of 895 as it stands today)

CC @Danil-Grigorev @soltysh @sttts This is the KCM part of the conversation we were having monday


JoelSpeed:
The first commit in the chain here is #450, only the second commit needs review, which adds a new external cloud volume plugin observer to configure KCM

JoelSpeed:
Switched bases to get GitHub to update the diff, was hoping it wouldn't trigger new test runs, sigh.

Anyway, this is good to go now, pending reviews

pkg/operator/configobservation/cloud/observe_cloud_volume_plugin.go
```diff
@@ -0,0 +1,110 @@
...
+	external, err := cloudprovider.IsCloudProviderExternal(infrastructure.Status.PlatformStatus, featureGate)
+	if err != nil {
+		return false, err
+	}
+
+	return external, nil
+}
+
+// getPlatformName returns the name of the cloud volume plugin if one is supported for the platform type.
+func getPlatformName(platformType configv1.PlatformType, recorder events.Recorder) string {
```
soltysh, line 84: Can't this be shared with library-go's version? Seems reasonable to expose that other one instead of copying

pkg/operator/configobservation/cloud/observe_cloud_volume_plugin.go
```diff
@@ -0,0 +1,110 @@
...
+
+	if !equality.Semantic.DeepEqual(prevObservedConfig, observedConfig) {
+		recorder.Event("ObserveCloudVolumePlugin", "observed change in config")
+	}
+
+	return observedConfig, errs
+}
+
+// isCloudProviderExternal determines whether or not the external cloud provider should be running within the cluster.
+func isCloudProviderExternal(listers configobservation.Listers, infrastructure *configv1.Infrastructure) (bool, error) {
```
soltysh, line 67: Same comment as below.

pkg/operator/configobservation/cloud/observe_cloud_volume_plugin.go
JoelSpeed: I was concerned about exposing those methods as eventually the config observer is going to go away, if we add it to library-go as a public method, we will eventually need to deprecate these right? 

I'll pull up a PR exposing those methods and we can discuss on that if there are concerns about the longevity of the methods

JoelSpeed: Created https://github.com/openshift/library-go/pull/1132, if you think that's appropriate to go ahead with, then I'm happy to update this one too

soltysh (approved):
/lgtm
/approve

PR #527, opened 2021-06-02
Bug 1954790: Use appropriate metric for PDB alerts

ravisantoshgudimetla:
As of now, we are using `kube_poddisruptionbudget_status_expected_pods`
metric to decide if the PDB is its limit or it exceeded where as
`kube_poddisruptionbudget_status_current_healthy` metric
gives the exact number of healthy replicas for a
deployment or other workload that can scale. So,
changing the alerts to use correct metric so that we
don't get unnecessary signals.

cc @soltysh @smarterclayton


soltysh (approved):
/lgtm
/approve

PR #529, opened 2021-06-07
Bug 1963730: kube-apiserver failed to load SNI cert and key

p0lyn0mial:
this PR renames `kube-controller-manager-certs` to `dynamic` **and** moves it under the target directory

After this PR:          /etc/kubernetes/static-pod-resources/kube-controller-manager-pod-REVISION/dynamic
Previously it was:  /etc/kubernetes/static-pod-resources/kube-controller-manager-certs

The target directory holds all files for the current static pod manifest.

As of today, the cert directory is accessed by at least two processes, namely the installer pod and the cert-syncer.
Up until now, there was no coordination between these processes that might have lead to many unexpected errors, like https://bugzilla.redhat.com/show_bug.cgi?id=1963730

Moving the cert directory to the target directory (revisioned) ensures that these processes won't step each other's toes


p0lyn0mial:
/hold

as it is not clear if the DR scripts will work

p0lyn0mial:
/close

closing in favour of https://github.com/openshift/cluster-kube-controller-manager-operator/pull/532

PR #538, opened 2021-06-22
Bug 1965562: recycler-for-nfs-... does not set requests or priorityClassName

dobsonj:
See the bug for details:
https://bugzilla.redhat.com/show_bug.cgi?id=1965562

`openshift-user-critical` seems ok to me, and the request sizes are the same as what we use on some of the CSI sidecars:
https://github.com/openshift/aws-ebs-csi-driver-operator/blob/master/assets/controller.yaml#L284-L287
This container is _very_ ephemeral (as in, typically can't catch it running without adding a sleep). It just removes files from an NFS volume after it's deleted, when `persistentVolumeReclaimPolicy: Recycle` is set on the PV.

@openshift/storage @wking : any other thoughts or feedback on the requests or priority class?

### DIFF
--- a/bindata/v4.1.0/kube-controller-manager/recycler-cm.yaml
+++ b/bindata/v4.1.0/kube-controller-manager/recycler-cm.yaml
@@ -29,5 +29,10 @@ data:
               name: vol
           securityContext:
             runAsUser: 0
+          priorityClassName: openshift-user-critical
+          resources:
+            requests:
+              memory: 50Mi
+              cpu: 10m
       volumes:
         - name: vol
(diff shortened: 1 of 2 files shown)


jsafrane:
I think `openshift-user-critical` is OK. In the worst case it will be deleted / pre-empted and KCM will start it again.

bertinatto:
/lgtm

@wking could you take a look?

wking (approved):

ravisantoshgudimetla (approved):
Thank you for the PR @dobsonj . The choice is always up to the developer. If you think it is ok for the recycler pod to be preempted and be killed, I am fine with it.


/lgtm

soltysh (approved):
/approve

PR #541, opened 2021-06-30
Refactor remaining controllers to factory

mfojtik:
So graceful shutdown work properly and context is wired properly.


soltysh (approved):
/lgtm

PR #544, opened 2021-07-02
Bump github.com/openshift/library-go

Fedosin:
New version contains changes related to external cloud provider featuregate handling.


Fedosin:
This change is already included in https://github.com/openshift/cluster-kube-controller-manager-operator/pull/450

PR #545, opened 2021-07-07
update cluster-policy-controller configuration

atiratree:
this is required for cluster-policy-controller refactoring: https://github.com/openshift/cluster-policy-controller/pull/65

- pass pod metadata and kubeconfig to the controller
- add new role for getting pods (required by event recorder)

### DIFF
--- a/bindata/assets/kube-controller-manager/pod.yaml
+++ b/bindata/assets/kube-controller-manager/pod.yaml
@@ -70,6 +70,15 @@ spec:
       initialDelaySeconds: 10
       timeoutSeconds: 10
   - name: cluster-policy-controller
+    env:
+      - name: POD_NAME
+        valueFrom:
+          fieldRef:
+            fieldPath: metadata.name
+      - name: POD_NAMESPACE
+        valueFrom:
+          fieldRef:
+            fieldPath: metadata.namespace
     image: ${CLUSTER_POLICY_CONTROLLER_IMAGE}
     imagePullPolicy: IfNotPresent
     terminationMessagePolicy: FallbackToLogsOnError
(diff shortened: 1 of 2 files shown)


bindata/v4.1.0/kube-controller-manager/cluster-policy-controller-role.yaml
```diff
@@ -0,0 +1,13 @@
+# This role is necessary for cluster-policy-controller to identify its own pod and owner for emitting events
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: openshift-kube-controller-manager
+  name: system:openshift:cluster-policy-controller
```
atiratree, line 6: is it okay to add a new service account?

atiratree, line 6: is it okay to add a new role/rolebinding?

atiratree: we will get the following warning unless we add the pod rule

```
W0708 12:14:57.795809       1 builder.go:209] unable to get owner reference (falling back to namespace): pods "kube-controller-manager-ip-10-0-168-17.eu-central-1.compute.internal" is forbidden: User "system:kube-controller-manager" cannot get resource "pods" in API group "" in the namespace "openshift-kube-controller-manager" 
```

The events will be emitted by the node (owner of the pod)

atiratree: removing, will be resolved by passing the owner reference directly

atiratree:
@soltysh I removed the additional RBAC and mirrored changes to the bindata/bootkube

PR #547, opened 2021-07-11
Start using embed module for assets

Fedosin:
This commit starts using "embed" module for static assets and removes the generated ones, created by go-bindata. This should help developers, as they don't have to regenerate bindata after each change in the assets.


soltysh (approved):
/lgtm
/approve

PR #552, opened 2021-07-23
Cloud Volume Observer: Extended arguments must be a slice

JoelSpeed:
Merging #525 appears to have broken the KCMO when it comes to setting this new flag

```
Operator degraded (TargetConfigController_SynchronizationError): TargetConfigControllerDegraded: "configmap": error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go struct field KubeControllerManagerConfig.extendedArguments of type v1.Arguments
```

Based on the other extended arguments observers, I've updated the observer here. Testing on a real cluster now, will report back shortly.


soltysh (approved):
/lgtm
/approve

JoelSpeed:
Have verified this is working on a cluster with the correct output in the extended arguments, and the pods seem to be coming up correctly
```
"external-cloud-volume-plugin": [
      "aws"
    ],
```

PR #555, opened 2021-07-26
Bug 1989073: Exclude openshift only CloudProvider feature gate from KCM config

JoelSpeed:
When KCM starts with a feature gate that it doesn't know, it logs an error. This PR should prevent that from happening by excluding the openshift only `ExternalCloudProvider` feature gate from the list of feature gates passed to KCM.

This relies on the functionality implemented in library-go to exclude blacklisted feature gates, as per https://github.com/openshift/library-go/blob/master/pkg/operator/configobserver/featuregates/observe_featuregates.go#L26


soltysh (approved):
/lgtm
/approve

PR #557, opened 2021-08-17
Bug 1989767: Set KCM leader election defaults

damemi:
This sets default LeaderElection values to better handle API server disruption as recommended in https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#handling-kube-apiserver-disruption


damemi (commented):
/cc @soltysh 
this is a little different than the KSO change https://github.com/openshift/cluster-kube-scheduler-operator/pull/365, because it doesn't look like KCM has componentconfig so we have to set the flags. Do we need to make the change to any bootkube assets too?

soltysh (approved):
/lgtm

deads2k:
This one in particular we should not change until graceful release becomes a thing for the KCM.

Without graceful release, the average case of leader changes is directly related to the lease duration.  Having a tight loop there is expensive, but the cost of waiting for two minutes for every normal release is too high to be reasonable.

The KCM is special in the sense that it is difficult to gracefully release and having a leader is critical to the smooth running of the cluster.  


https://github.com/kubernetes/kubernetes/pull/101379 and https://github.com/kubernetes/kubernetes/pull/101125 are in progress to make the process gracefully release

/hold until the KCM can gracefully release the lease.

PR #558, opened 2021-08-20
Set operator condition that it doesn't own cloud provider

Fedosin:
When external cloud provider is applied, the operator should set the condition that it doesn't own this component anymore.

It will help migrating the system from in-tree to external cloud providers:
https://github.com/openshift/enhancements/blob/master/enhancements/machine-api/out-of-tree-provider-support.md#migration-procedure

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -31,6 +31,7 @@ import (
 	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/openshift/library-go/pkg/operator/condition"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/management"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
@@ -296,6 +297,10 @@ func createTargetConfigController(ctx context.Context, syncCtx factory.SyncConte
 		return true, nil
 	}
 
+	if err = setCloudControllerOwnerCondition(c); err != nil {
+		return true, err
+	}
+
 	condition := operatorv1.OperatorCondition{
 		Type:   "TargetConfigControllerDegraded",
 		Status: operatorv1.ConditionFalse,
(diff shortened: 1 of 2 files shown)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -304,6 +304,39 @@ func createTargetConfigController(ctx context.Context, syncCtx factory.SyncConte
...
+	}
+	if externalCloudProviderApplied {
+		cond := operatorv1.OperatorCondition{
+			Type:   "CloudControllerOwner",
+			Status: operatorv1.ConditionFalse,
+		}
+		if _, _, err := v1helpers.UpdateStaticPodStatus(c.operatorClient, v1helpers.UpdateStaticPodConditionFn(cond)); err != nil {
+			return true, err
+		}
+	}
```
JoelSpeed, line 320: Does this wait until the changes have rolled out before applying the condition?

Fedosin: I added a check to make sure that static pods are available before setting the condition.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -296,6 +297,37 @@ func createTargetConfigController(ctx context.Context, syncCtx factory.SyncConte
...
+		// Once we have successfully set CloudControllerOwner to False, we should not change it again.
+		if !v1helpers.IsOperatorConditionFalse(status.Conditions, "CloudControllerOwner") {
+			cond := operatorv1.OperatorCondition{
+				Type:   "CloudControllerOwner",
+				Status: operatorv1.ConditionTrue,
+			}
+
+			// We set the condition to False only if at least one static pod has successfully started. After that
+			// we are sure that KCM doesn't own CloudController anymore. Otherwise we set it to True by implying
+			// that KCM is still the owner.
```
JoelSpeed, line 320: Why does one static pod starting mean that KCM doesn't own it anymore? Should we not wait for them all to be updated to the new version?

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -307,6 +339,24 @@ func createTargetConfigController(ctx context.Context, syncCtx factory.SyncConte
...
+	}
+
+	cloudProvider, found, err := unstructured.NestedSlice(observedConfig, "extendedArguments", "cloud-provider")
+	if !found || err != nil {
+		return false, err
+	}
+
+	if len(cloudProvider) == 1 && (cloudProvider[0] == "external" || cloudProvider[0] == "") {
+		return true, nil
+	}
```
JoelSpeed, line 355: `cloud-provider == ""` is not the same as being external. Does this imply that on a platform none or other platform that doesn't currently have in-tree, we are going to add the condition to say KCM doesn't own a cloud controller, even though no cloud controller exists?

At first glance I can't see any harm in this, just wasn't expecting that 🤔

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
Fedosin: Now we check `NodeInstallerProgressingCondition` condition to make sure that all pods are at the latest revision

JoelSpeed:
Changes make sense to me

@Fedosin please can you make sure you've manually tested this and post the results of the testing as a comment here

/lgtm

JoelSpeed (commented):
I like the way the condition stuff has been extracted, I think we can make it a bit easier to read and follow though if we reverse some of the conditions and unindent some of the code.

Would be great to see some unit tests for the `setCloudControllerOwnerCondition` as well to make sure it's doing what we expect it to do

JoelSpeed:
Code changes are looking much cleaner now, any thoughts about how to unit test this code?

Fedosin:
yep, I'm working on unit tests

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -307,6 +312,69 @@ func createTargetConfigController(ctx context.Context, syncCtx factory.SyncConte
...
+	// Once we have successfully set CloudControllerOwner to False, we should not change it again.
+	if v1helpers.IsOperatorConditionFalse(status.Conditions, "CloudControllerOwner") {
+		return nil
+	}
+
+	// Get a config map for the latest revision. Then check if it contains "cloud-provider" argument which is equal
+	// to "external" or just empty.
+	latestRevisionConfigMapName := fmt.Sprintf("config-%d", status.LatestAvailableRevision)
+	cm, err := c.configMapLister.ConfigMaps(operatorclient.TargetNamespace).Get(latestRevisionConfigMapName)
```
soltysh, line 331: You should only reach for `config` config map, which will always reflect the latest rolled out revision content.

soltysh (changes requested):

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
Fedosin: Actually that was done intentionally to prevent a race condition in KCMO. Imagine a situation:
There is revision 10 which doesn't have cloud-provider set to external. KCM object status has "10" in latestAvailableRevision. NodeInstallerProgressingCondition is false, the message is "AllNodesAtLatestRevision".

Then we apply the FeatureGate to enable CCMs. New revision "11" is created. "config" cm starts to point to "config-11". But latestAvailableRevision for KCM is still "10", and the condition doesn't change for some short period of time. A second later the controller will update latestAvailableRevision to "11", but it may be too late, because we've already set CloudControllerOwner to False.

So, if we fetch just "config", we may mistakenly think that the right configuration has already been applied, because KCMO will report AllNodesAtLatestRevision, which might not be true.

To avoid this race condition we take into consideration the LatestAvailableRevision value and fetch the config map with this number as a suffix.

soltysh (approved):
/lgtm
/approve

PR #565, opened 2021-09-22
podsecurity: enforce privileged for openshift-kube-controller-manager namespace

s-urbaniak:
Starting with OpenShift 4.10 we are introducing PodSecurity admission (https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/2579-psp-replacement).

Currently, all pods are marked as privileged, however, over time we want to enforce at least baseline, admirably restricted as default. In order not to break control plane workloads this allows workloads in `openshift-kube-controller-manager` namespace to run privileged pods.

See https://github.com/openshift/enhancements/pull/899 for more details (and excuse the eventual consistency of updates).

/cc @deads2k @sttts


soltysh (approved):
/lgtm
/approve

PR #570, opened 2021-11-09
introduce creation of openshift-infra ns and recycler SA

atiratree:
Moved this logic from openshift/kubernetes to collocate pv-recycler-controller SA with recycler-config/recycler-pod.

But I am not sure if we should move the `openshift-infra` namespace creation to here since it is also used for resource creation in other components: `openshift-controller-manager`, `cluster-openshift-controller-manager-operator`, `openshift-apiserver`. 
Shouldn't we move the creation of this namespace to cluster-version-operator as the responsibility of its creation probably does not belong to KCM operator?

For now I have given it a higher priority `0000_00` so its creation preceeds all the relevant components. It is also beneficial for `cluster-policy-controller` as it uses this namespace.

With lower priority `00_` I am observing creation of the namespace without the `workload.openshift.io/allowed: management` annotation until KCM operator reconciles it. I don't know why this is happening. 
- I could not locate any other component that would create this openshift-infra namespace.
- I have updated all the relevant images in my release image (openshift-apiserver, hyperkube and cluster-kube-controller-manager-operator). Maybe there is some mistake in my building process?
- Is there something that could strip the namespace labels?

relevant PRs that remove creation of openshfit-infra namespace:
- https://github.com/openshift/kubernetes/pull/1039
- https://github.com/openshift/openshift-apiserver/pull/258


soltysh (approved):
/lgtm
/approve
/retest

PR #577, opened 2021-11-24
Collect kube-system namespace as a related resource to kube-controller-manager operator

xueqzhan:
Add kube-system namespace to relatedobjects

We noticed that the operator is using a configmap for leader election that is not a related resource and thus is not gathered by must-gather. Adding the namespace as a related resource will correct this.

### DIFF
--- a/manifests/0000_25_kube-controller-manager-operator_07_clusteroperator.yaml
+++ b/manifests/0000_25_kube-controller-manager-operator_07_clusteroperator.yaml
@@ -32,6 +32,9 @@ status:
       name: openshift-kube-controller-manager-operator
       resource: namespaces
     - group: ""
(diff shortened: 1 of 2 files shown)


dgoodwin:
@xueqzhan Suggest adding a little info to the commit msg, specifically that we discovered the configmap used for the kube-controller-manager lives in this namespace, and isn't getting picked up by must-gather.

xueqzhan:
@dgoodwin Do you mean commit comment or PR comment? Will that be a little too long for commit comment?

dgoodwin:
Too long for the first line of a commit comment, but totally acceptable for the body of a commit comment.

soltysh (approved):
/lgtm
/approve

PR #583, opened 2021-12-17
pdb: add possibility to silence the PodDisruptionBudgetAtLimit alert

acardace:
When the label 'alerts.openshift.io/PodDisruptionBudgetAtLimit:disabled' is present on a pdb the alert won't fire for that resource.

Signed-off-by: Antonio Cardace <acardace@redhat.com>

### DIFF
--- a/manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
+++ b/manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
@@ -28,7 +28,7 @@ spec:
             description: The pod disruption budget is at the minimum disruptions allowed level. The number of current healthy pods is equal to the desired healthy pods.
             runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-kube-controller-manager-operator/PodDisruptionBudgetAtLimit.md
           expr: |
-            max by(namespace, poddisruptionbudget) (kube_poddisruptionbudget_status_current_healthy == kube_poddisruptionbudget_status_desired_healthy and on (namespace, poddisruptionbudget) kube_poddisruptionbudget_status_expected_pods > 0)
+            max by(namespace, poddisruptionbudget) (kube_poddisruptionbudget_status_current_healthy == kube_poddisruptionbudget_status_desired_healthy and on (namespace, poddisruptionbudget) kube_poddisruptionbudget_status_expected_pods > 0 unless on (namespace, poddisruptionbudget) kube_poddisruptionbudget_labels{label_alerts_openshift_io_pod_disruption_budget_at_limit="disabled"} == 1)
           for: 60m
           labels:
             severity: warning


acardace:
@arajkumar can anyone take a look at this?

arajkumar:
Alert expression looks good to me.

fabiand:
@soltysh can you help getting this reviewed?

Currently some PDB alerst around VMs are useless/annoying, we need this fix to get rid of it.

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
```diff
@@ -26,7 +26,7 @@ spec:
             summary: The pod disruption budget is preventing further disruption to pods.
             description: The pod disruption budget is at minimum disruptions allowed level. The number of current healthy pods is equal to desired healthy pods.
           expr: |
-            max by(namespace, poddisruptionbudget) (kube_poddisruptionbudget_status_current_healthy == kube_poddisruptionbudget_status_desired_healthy)
+            max by(namespace, poddisruptionbudget) (kube_poddisruptionbudget_status_current_healthy == kube_poddisruptionbudget_status_desired_healthy unless on (namespace, poddisruptionbudget) kube_poddisruptionbudget_labels{label_alerts_openshift_io_pod_disruption_budget_at_limit="disabled"} == 1)
```
ravisantoshgudimetla, line 29: Why is this needed? Some background would help. We were also thinking of splitting this alert into 2 alerts one for platform and other for customer namespaces.

acardace: Hi @ravisantoshgudimetla, basically in KubeVirt we have PDBs for VMs that have the number of pods which is equal to the minAvailable parameter at all times, this is by design so that we can live-migrate VMs instead of having k8s evicting them.
 
This alert in that context is just noise that might confuse a cluster-admin and that's why we'd like a way to opt-out from this behavior.

acardace:
@ravisantoshgudimetla @mfojtik  can anyone look at this?

david-martin:
I believe there's a use case for disabling this alert in clusters that use strimzi (kafka) and the strimzi drain cleaner https://strimzi.io/blog/2021/09/24/drain-cleaner/
The PDB is deliberately set at a value that could cause an alert to fire.
However it's within the control of the operator to manage pods during node drain cases e.g. node upgrades.

acardace:
@ravisantoshgudimetla @mfojtik can anyone review this?

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
soltysh: @ravisantoshgudimetla i don't think that the split will help in this particular situation since KubeVirt bits are still part of platform and not customer. My only worry is that this might be overused by users to silence their alerts.

fabiand: Why would it be concerning if a user used this label as well?

Isn't it actually better if a user uses this, instead of an admin?

I suspect that there are other alerts who would fire in case of i.e. a drain being blocked

atiratree: > My only worry is that this might be overused by users to silence their alerts.

one option would be to hardcode this specifically for this use case and filter on a PDB name: `poddisruptionbudget=~"kubevirt-disruption-budget-.*"`. But this approach could bring higher maintenance if similar requests come in the future.

stu-gott:
This is still something we're very interested in. There doesn't appear to be any good alternative.

fabiand:
@atiratree review would be appreciated

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
fabiand: As we see in the comments, kafka is also impacted by this alert.

stu-gott:
@ravisantoshgudimetla are you still the correct person to review this? We still need this functionality.

atiratree:
@stu-gott I am the contact person for this issue. I am trying pursue an enhancement that would solve this in https://github.com/openshift/enhancements/pull/1328

atiratree:
Yes, we are still evaluating the alternatives in the enhancements proposal.

atiratree:
After the recent discussion we decided not go forward with this solution and go instead with the silencing mechanism. Referencing a KEP with more details regarding this: https://github.com/openshift/enhancements/pull/1328#issuecomment-1429790069 

@acardace can you please close this PR? If needed you can reopen the KEP and start a new disscussion

PR #589, opened 2022-01-13
Bug 2040132: drop --port flag

soltysh:
/assign @ingvagabund


ingvagabund:
https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/pkg/cmd/render/render_test.go#L187 needs to be dropped as well

PR #594, opened 2022-01-28
Bug 2045872: allow cluster-policy-controller to fallback to default cert

atiratree:
- default self-signed certificate is used when service-ca controller is
  lagging or unable to start
- kcm pod will be redeployed and the certificate changed as soon as serving-cert is created by
  service-ca

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -23,6 +23,7 @@ import (
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/cert"
+	"k8s.io/klog/v2"
 
 	kubecontrolplanev1 "github.com/openshift/api/kubecontrolplane/v1"
 	openshiftcontrolplanev1 "github.com/openshift/api/openshiftcontrolplane/v1"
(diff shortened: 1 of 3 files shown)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -391,17 +396,46 @@ func manageKubeControllerManagerConfig(ctx context.Context, client corev1client.
...
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, false, err
+	} else if apierrors.IsNotFound(err) {
+		// Should only apply when starting the cluster so cluster-policy-controller is able to annotate openshift-service-ca namespace.
+		// Then service-ca controller should start and create serving-cert.
+		// We will put the serving-cert into the config as soon as it appears which will then trigger new installer.
+
+		klog.V(1).Info("serving-cert not found: falling back to default self-signed certificate in cluster-policy-controller")
+		configOverride := "{\"servingInfo\": { \"certFile\": \"\", \"keyFile\": \"\"} }"
+		configYamls = append(configYamls, []byte(configOverride))
```
atiratree, line 428: to trigger defaulting here: https://github.com/openshift/library-go/blob/master/pkg/controller/controllercmd/cmd.go#L181

atiratree:
info about serving-cert secret:

serving-cert secret is created by service-ca controller from here https://github.com/openshift/cluster-kube-controller-manager-operator/blob/c9a1b00ccb2cd7b9d8d46da790f3b519ae336e89/bindata/assets/kube-controller-manager/svc.yaml#L7
Then the [serving-cert](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/pkg/operator/starter.go#L248) secret name is passed to the installer pod here  https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/installer/installer_controller.go#L890.

Then the  installer pod copies the secret to the static pod: https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/installerpod/cmd.go#L297

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
soltysh: would be handy to have that link/location to that defaulting in comment.

soltysh (approved):
/lgtm
/approve

soltysh (approved):
/lgtm

PR #613, opened 2022-04-14
Bug 1918690: update resource-graph to include current resources

atiratree:
### After

after

### Before

before


soltysh (approved):
/lgtm

PR #615, opened 2022-04-22
introduce a reverse proxy to KCM and CPC in order to enable CPC metrics 

atiratree:
- KCM is now serving on port 10358, so the proxy can take KCM's former port 10257 which is opened on nodes and can be used by prometheus for metrics

- serving-cert is used by the proxy and renamed to kcm-cpc-reverse-proxy-serving-cert
- introduce new localhost certificates
  - cpc-localhost-serving-cert
  - kcm-localhost-serving-cert

### DIFF
--- a/bindata/assets/config/defaultconfig.yaml
+++ b/bindata/assets/config/defaultconfig.yaml
@@ -29,7 +29,7 @@ extendedArguments:
   cluster-signing-duration:
   - "720h"
   secure-port:
-  - "10257"
+  - "10358"
   cert-dir:
   - "/var/run/kubernetes"
   root-ca-file:
--- a/bindata/assets/kube-controller-manager/svc.yaml
+++ b/bindata/assets/kube-controller-manager/svc.yaml
@@ -4,7 +4,7 @@ metadata:
   namespace: openshift-kube-controller-manager
   name: kube-controller-manager
   annotations:
-    service.beta.openshift.io/serving-cert-secret-name: serving-cert
+    service.beta.openshift.io/serving-cert-secret-name: kcm-cpc-reverse-proxy-serving-cert
   labels:
     prometheus: "kube-controller-manager"
 spec:
--- a/manifests/0000_90_kube-controller-manager-operator_04_servicemonitor-controller-manager.yaml
+++ b/manifests/0000_90_kube-controller-manager-operator_04_servicemonitor-controller-manager.yaml
@@ -78,3 +78,46 @@ spec:
   selector:
     matchLabels:
       prometheus: kube-controller-manager
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    k8s-app: kube-controller-manager
+  name: cluster-policy-controller
+  namespace: openshift-kube-controller-manager
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+spec:
+  endpoints:
+    - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      interval: 30s
+      metricRelabelings:
+        - action: drop
+          regex: etcd_(debugging|disk|request|server).*
+          sourceLabels:
+            - __name__
+        - action: drop
+          regex: rest_client_request_latency_seconds_(bucket|count|sum)
+          sourceLabels:
+            - __name__
+        - action: drop
+          regex: root_ca_cert_publisher_sync_duration_seconds_(bucket|count|sum)
+          sourceLabels:
+            - __name__
+      port: https
+      scheme: https
+      path: /cluster-policy-controller/metrics
+      tlsConfig:
+        caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+        serverName: kube-controller-manager.openshift-kube-controller-manager.svc
+        certFile: /etc/prometheus/secrets/metrics-client-certs/tls.crt
+        keyFile: /etc/prometheus/secrets/metrics-client-certs/tls.key
+  namespaceSelector:
+    matchNames:
+      - openshift-kube-controller-manager
+  selector:
+    matchLabels:
+      prometheus: kube-controller-manager
--- a/manifests/0000_90_kube-controller-manager-operator_04_servicemonitor-controller-manager.yaml
+++ b/manifests/0000_90_kube-controller-manager-operator_04_servicemonitor-controller-manager.yaml
@@ -65,7 +65,7 @@ spec:
       sourceLabels:
       - __name__
     - action: drop
-      regex: rest_client_request_latency_seconds_(bucket|count|sum)
+      regex: rest_client_request_duration_seconds_(bucket|count|sum)
       sourceLabels:
       - __name__
     - action: drop
@@ -114,7 +114,7 @@ spec:
           sourceLabels:
             - __name__
         - action: drop
-          regex: rest_client_request_latency_seconds_(bucket|count|sum)
+          regex: rest_client_request_duration_seconds_(bucket|count|sum)
           sourceLabels:
             - __name__
         - action: drop
--- a/manifests/0000_90_kube-controller-manager-operator_04_servicemonitor-controller-manager.yaml
+++ b/manifests/0000_90_kube-controller-manager-operator_04_servicemonitor-controller-manager.yaml
@@ -52,6 +52,13 @@ spec:
   endpoints:
   - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
     interval: 30s
+    relabelings:
+      - action: replace
+        targetLabel: container
+        replacement: kube-controller-manager
+      - action: replace
+        targetLabel: job
+        replacement: kube-controller-manager
     metricRelabelings:
     - action: drop
       regex: etcd_(debugging|disk|request|server).*
@@ -94,6 +101,13 @@ spec:
   endpoints:
     - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
       interval: 30s
+      relabelings:
+        - action: replace
+          targetLabel: container
+          replacement: cluster-policy-controller
+        - action: replace
+          targetLabel: job
+          replacement: cluster-policy-controller
       metricRelabelings:
         - action: drop
           regex: etcd_(debugging|disk|request|server).*
--- a/pkg/cmd/reverseproxy/cmd.go
+++ b/pkg/cmd/reverseproxy/cmd.go
@@ -1,0 +1,227 @@
+package reverseproxy
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/library-go/pkg/crypto"
+)
+
+// reverseProxyOpts holds values to run the proxy.
+type reverseProxyOpts struct {
+	bindAddress string
+	certFile    string
+	keyFile     string
+
+	kcmPort         string
+	kcmCABundleFile string
+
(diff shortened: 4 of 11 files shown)


bindata/assets/kube-controller-manager/svc.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   namespace: openshift-kube-controller-manager
   name: kube-controller-manager
   annotations:
-    service.beta.openshift.io/serving-cert-secret-name: serving-cert
+    service.beta.openshift.io/serving-cert-secret-name: kcm-cpc-reverse-proxy-serving-cert
```
atiratree, line 7: we have to change the name of the secret/file because cluster-policy-controller searches this location by default.

But it is causing small hiccups when upgrading. The in between pods crash before the final revision is reached as the moving of the secret is asynchronous.

so we will get a few of these before the final revision is reached

> open /etc/kubernetes/static-pod-resources/secrets/serving-cert/tls.crt: no such file or directory

pkg/operator/starter.go
```diff
@@ -238,14 +238,22 @@ var deploymentConfigMaps = []revision.RevisionResource{
...
+	{Name: "cpc-localhost-serving-ca"},
 }
 
 // deploymentSecrets is a list of secrets that are directly copied for the current values.  A different actor/controller modifies these.
 var deploymentSecrets = []revision.RevisionResource{
 	{Name: "service-account-private-key"},
 
+	// created by certrotation controller
+	{Name: "kcm-localhost-serving-cert"},
+	{Name: "cpc-localhost-serving-cert"},
```
atiratree, line 253: we may get an err/warnings for a while in the operator before the certrorationcontroller creates the resources. This might take a few seconds and generate 10s of errors. Is that okay?

```
E0421 17:51:38.100472       1 base_controller.go:270] "InstallerController" controller failed to sync "key", err: missing required resources: [secrets: csr-signer,kube-controller-manager-client-cert-key, configmaps: cpc-localhost-serving-ca-7,kcm-localhost-serving-ca-7, secrets: cpc-localhost-serving-cert-7,kcm-localhost-serving-cert-7,localhost-recovery-client-token-7,service-account-private-key-7]
I0421 17:51:38.261844       1 event.go:282] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-kube-controller-manager-operator", Name:"kube-controller-manager-operator", UID:"19fa4dd8-0702-4c69-8768-6a6748af6258", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Warning' reason: 'RequiredInstallerResourcesMissing' secrets: csr-signer,kube-controller-manager-client-cert-key, configmaps: cpc-localhost-serving-ca-7,kcm-localhost-serving-ca-7, secrets: cpc-localhost-serving-cert-7,kcm-localhost-serving-cert-7,localhost-recovery-client-token-7,service-account-private-key-7
```

bindata/assets/config/defaultconfig.yaml
```diff
@@ -29,7 +29,7 @@ extendedArguments:
   cluster-signing-duration:
   - "720h"
   secure-port:
-  - "10257"
+  - "10358"
```
atiratree, line 32: since prometheus is not running in bootstrap phase it probably doesn't make sense to run the proxy there. One thing that is a bit off is that the kube-controller-manager still has the port 10257 in a bootstrap phase and after that proxy overtakes it

manifests/0000_90_kube-controller-manager-operator_04_servicemonitor-controller-manager.yaml
```diff
@@ -78,3 +78,46 @@ spec:
...
+  endpoints:
+    - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      interval: 30s
+      metricRelabelings:
+        - action: drop
+          regex: etcd_(debugging|disk|request|server).*
+          sourceLabels:
+            - __name__
+        - action: drop
+          regex: rest_client_request_latency_seconds_(bucket|count|sum)
```
aojea, line 117: I don't know when this has changed , but is `rest_client_request_duration_seconds` IIUIC https://github.com/kubernetes/kubernetes/blob/17381762dd6caafe72cda9e9c39e42405391f1e9/staging/src/k8s.io/component-base/metrics/prometheus/restclient/metrics.go#L36

manifests/0000_90_kube-controller-manager-operator_04_servicemonitor-controller-manager.yaml
```diff
@@ -78,3 +78,46 @@ spec:
...
+          sourceLabels:
+            - __name__
+        - action: drop
+          regex: root_ca_cert_publisher_sync_duration_seconds_(bucket|count|sum)
+          sourceLabels:
+            - __name__
+      port: https
+      scheme: https
+      path: /cluster-policy-controller/metrics
```
aojea, line 126: so, basically, you add a reverse proxy so you multiplex metrics by url

https://host:port/metrics -> kcm
https://host:port/cluster-policy/cluster-policy-controller/metrics -> cpc 

I'm not familiar with Openshift, is this the common way to do this in Openshift, is an interesting case indeed?

bindata/assets/config/defaultconfig.yaml
aojea: yeah, 10257 is the default port, can there be more consequences with this change?


```
--secure-port int     Default: 10257
```

aojea: by the way, where do you set the `--secure-port` option in kcm to listen on port 10358?

aojea: oh, it is generated when it renders based on the manifest?

pkg/cmd/reverseproxy/cmd.go
```diff
@@ -0,0 +1,225 @@
...
+	kcmURL, err := url.Parse("https://" + net.JoinHostPort("127.0.0.1", r.kcmPort))
+	if err != nil {
+		return err
+	}
+	cpcURL, err := url.Parse("https://" + net.JoinHostPort("127.0.0.1", r.cpcPort))
+	if err != nil {
+		return err
+	}
+
+	kcmProxy, err := newReverseProxy(kcmURL, r.kcmCABundleFile, "")
```
aojea, line 122: I always  have doubts with this, it should be empty or just a single slash "/"?

pkg/cmd/reverseproxy/cmd.go
```diff
@@ -0,0 +1,225 @@
...
+	if len(r.cpcPathPrefix) == 0 {
+		return errors.New("missing required flag: --cpc-url-path-prefix")
+	}
+	return nil
+}
+
+// Run contains the logic of the proxy command.
+func (r *reverseProxyOpts) Run() error {
+	if len(r.certFile) == 0 || len(r.keyFile) == 0 {
+		err := r.generateSelfSignedCert()
```
soltysh, line 107: Why not fail instead of generating self-signed certs?

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -118,7 +118,86 @@ func newCertRotationController(
...
+			Informer:      kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps(),
+			Lister:        kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps().Lister(),
+			Client:        configMapsGetter,
+			EventRecorder: eventRecorder,
+		},
+		certrotation.RotatedSelfSignedCertKeySecret{
+			Namespace:              operatorclient.TargetNamespace,
+			Name:                   "kcm-localhost-serving-cert",
+			Validity:               850 * rotationDay, // 28 months - kcm-cpc-reverse-proxy-serving-cert has 26 months
+			Refresh:                425 * rotationDay, // 14 months - kcm-cpc-reverse-proxy-serving-cert has 13 months
```
soltysh, line 148: Any particular reason for this long rotation? What is the problem if we rotate these like all the remaining serving certs every 60/30 days? /cc @deads2k I'd like your input on this matter

soltysh: This applies to other certs as well.

soltysh: The only reason the localhost we have in kas-o are 10y validity is they are meant for recovery, but this one should be fine rotating like all the others with 60 validity and 30 refresh, imo.

soltysh (commented):
The biggest unknown to me is the certs validity period, that's why I'd like to have confirmation from @deads2k about it.

manifests/0000_90_kube-controller-manager-operator_04_servicemonitor-controller-manager.yaml
atiratree: thanks for noticing, yes it seems to have been removed 2 years ago https://github.com/kubernetes/kubernetes/commit/606e4503cf4cea9f05cfd467b88c88cf9c5648d9

I have added `rest_client_request_duration_seconds` to be dropped

manifests/0000_90_kube-controller-manager-operator_04_servicemonitor-controller-manager.yaml
atiratree: Yes. But I am not sure how common (if anybody even does this approach) this is in openshift. 

We decided to use this approach after a discussion with installer people: apparently there is no process/solution to update (open/close) new ports on our nodes (usually via cloud provider API) and we do it on install only. And it seems there is no plan for such a solution in the future. This is not a problem for new clean installs, but we are unable to open a port for cluster-policy-controller (10357) on upgrades.

The proxy seems the most feasible solution how to achieve scrapping of our metrics.

aojea: It was mostly curiosity ,  agree the proxy seems the simplest and easiest solution to this problem

pkg/cmd/reverseproxy/cmd.go
atiratree: I just started with `""` for simplicity. But I guess `"/"` looks better since we can reuse the same const variable also for registering the handle func. Updated

pkg/cmd/reverseproxy/cmd.go
atiratree: I thought it would be preferable in case something goes wrong with the cert / secret / cluster-ca operator. 

Even though the metrics collection would stop working and we would receive the alerts, it might be advantageous to have access to kcm even in such cases (for example for debugging eg. `debug/controllers/garbagecollector/graph` or for manually fetching the metrics).

Also I just found out it might be needed for this use case https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/core/rest/storage_core.go#L396 but I am not sure how necessary is it.

pkg/operator/certrotationcontroller/certrotationcontroller.go
atiratree: Since the front serving cert (kube-controller-manager svc)  of the proxy rotates after one year (service-ca operator), I thought  it might not be necessary to rotate the localhost certs faster. But I am not against changing it to any other value if we want to follow conventions.

pkg/cmd/reverseproxy/cmd.go
```diff
@@ -0,0 +1,227 @@
...
+
+	proxy.Director = func(req *http.Request) {
+		req.URL.Scheme = target.Scheme
+		req.URL.Host = target.Host
+		if len(pathPrefix) != 0 && pathPrefix != "/" {
+			req.URL.Path = strings.TrimPrefix(req.URL.Path, pathPrefix)
+			if len(req.URL.RawPath) != 0 {
+				req.URL.RawPath = strings.TrimPrefix(req.URL.EscapedPath(), pathPrefix)
+			}
+		}
```
aojea, line 213: this will need some tests to verify that the pathPrefix and Trim doesn't do something we don't want

I will put it in its own function and unit test it
```go
func trimPathPrefix(pathPrefix string, path string) string {
}
```

atiratree: good idea, I will add this if we decide to proceed with this PR as there are currently concerns a proxy is not a safe approach.

aojea: that is why I've asked https://github.com/openshift/cluster-kube-controller-manager-operator/pull/615#discussion_r855843178 , I think this will be a common problem that will need a standard solution @deads2k @soltysh

atiratree:
/hold

this needs to be reworked as there are issues with using a proxy

PR #618, opened 2022-04-25
policy-controller RBAC: use the new leases API

stlaz:
Use the new API once the cluster-policy-controller gets an updated library-go.

/assign @soltysh 
/cc @tkashem


soltysh (approved):
/lgtm
/approve

PR #580, opened 2021-12-06
add alerts for garbage collector

atiratree:
build on top of https://github.com/kubernetes/kubernetes/pull/106844 and adds warning alerts when garbage collector cannot reconcile

PR #623, opened 2022-05-11
Bug 2099668: introduce GC Watcher controller and add alerts for GarbageCollector

atiratree:
- watches important Garbage Collector alerts and degrades KCMo if they
  happen
- also checks for the presence of these alerts in prometheus

supersedes #580

### DIFF
--- a/manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
+++ b/manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
@@ -39,3 +39,12 @@ spec:
           for: 15m
           labels:
             severity: critical
+        - alert: GarbageCollectorSyncFailed
+          annotations:
+            summary: There was a problem with syncing the resources for garbage collection.
+            description: Garbage Collector had a problem with syncing and monitoring the available resources. Please see KubeControllerManager logs for more details.
+          expr: |
+            rate(garbagecollector_controller_resources_sync_error_total{}[5m]) > 0
+          for: 60m
+          labels:
+            severity: warning
--- a/pkg/operator/gcwatchercontroller/prometheus_client.go
+++ b/pkg/operator/gcwatchercontroller/prometheus_client.go
@@ -1,0 +1,63 @@
+package gcwatchercontroller
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"time"
+
+	prometheusapi "github.com/prometheus/client_golang/api"
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/transport"
+
+	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/operatorclient"
+)
+
+func newPrometheusClient(ctx context.Context, configMapClient corev1client.ConfigMapsGetter) (prometheusv1.API, error) {
+	host := "thanos-querier.openshift-monitoring.svc"
+
+	saToken, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/token")
+	if err != nil {
+		return nil, fmt.Errorf("error reading service account token: %w", err)
+	}
+
+	routerCAConfigMap, err := configMapClient.ConfigMaps(operatorclient.GlobalMachineSpecifiedConfigNamespace).Get(ctx, "service-ca", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	bundlePEM := []byte(routerCAConfigMap.Data["ca-bundle.crt"])
+
+	roots := x509.NewCertPool()
+	roots.AppendCertsFromPEM(bundlePEM)
+
+	client, err := prometheusapi.NewClient(prometheusapi.Config{
+		Address: "https://" + net.JoinHostPort(host, "9091"),
+		RoundTripper: transport.NewBearerAuthRoundTripper(
+			string(saToken),
+			&http.Transport{
+				Proxy: http.ProxyFromEnvironment,
+				DialContext: (&net.Dialer{
+					Timeout:   30 * time.Second,
+					KeepAlive: 30 * time.Second,
+				}).DialContext,
+				TLSHandshakeTimeout: 10 * time.Second,
+				TLSClientConfig: &tls.Config{
+					RootCAs:    roots,
+					ServerName: host,
+				},
+			},
+		),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return prometheusv1.NewAPI(client), nil
+}
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -15,6 +15,7 @@ import (
 	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/certrotationcontroller"
 	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/configobservation/configobservercontroller"
 	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/configobservation/node"
+	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/gcwatchercontroller"
 	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/operatorclient"
 	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/resourcesynccontroller"
 	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/targetconfigcontroller"
@@ -261,6 +262,10 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 		cc.EventRecorder,
 	)
 
+	gcWatcherController := gcwatchercontroller.NewGarbageCollectorWatcherController(operatorClient, kubeInformersForNamespaces, kubeClient, cc.EventRecorder, []string{
+		"GarbageCollectorSyncFailed",
+	})
+
 	configInformers.Start(ctx.Done())
 	kubeInformersForNamespaces.Start(ctx.Done())
 	dynamicInformers.Start(ctx.Done())
@@ -275,6 +280,7 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 	go saTokenController.Run(ctx, 1)
 	go staleConditions.Run(ctx, 1)
 	go latencyProfileController.Run(ctx, 1)
+	go gcWatcherController.Run(ctx, 1)
 
 	<-ctx.Done()
 	return nil
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/openshift/build-machinery-go v0.0.0-20211213093930-7e33a7eb4ce3
 	github.com/openshift/client-go v0.0.0-20220525160904-9e1acff93e4a
 	github.com/openshift/library-go v0.0.0-20220622115547-84d884f4c9f6
+	github.com/prometheus/client_golang v1.12.1
 	github.com/prometheus/common v0.32.1
 	github.com/spf13/cobra v1.4.0
 	github.com/spf13/pflag v1.0.5
@@ -66,7 +67,6 @@ require (
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pkg/profile v1.3.0 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	github.com/prometheus/client_golang v1.12.1 // indirect
 	github.com/prometheus/client_model v0.2.0 // indirect
 	github.com/prometheus/procfs v0.7.3 // indirect
 	github.com/robfig/cron v1.2.0 // indirect
--- a/pkg/operator/gcwatchercontroller/fake_prometheus_client_test.go
+++ b/pkg/operator/gcwatchercontroller/fake_prometheus_client_test.go
@@ -1,0 +1,116 @@
+package gcwatchercontroller
+
+import (
+	"context"
+	"time"
+
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	prometheusmodel "github.com/prometheus/common/model"
+)
+
+func newFakePrometheusClient(firingAlerts []string, responseError error) prometheusv1.API {
+	queryResult := prometheusmodel.Vector{}
+
+	for _, alert := range firingAlerts {
+		alertLabels := prometheusmodel.LabelSet{
+			prometheusmodel.AlertNameLabel: prometheusmodel.LabelValue(alert),
+		}
+		smpl := &prometheusmodel.Sample{
+			Metric: prometheusmodel.Metric(alertLabels),
+		}
+		queryResult = append(queryResult, smpl)
+	}
+
+	client := fakePrometheusClient{}
+	client.queryResultVal = queryResult
+	client.queryErr = responseError
+	return client
+}
+
+type fakePrometheusClient struct {
+	queryResultVal prometheusmodel.Vector
+	queryErr       error
+}
+
+func (f fakePrometheusClient) Query(ctx context.Context, query string, ts time.Time) (prometheusmodel.Value, prometheusv1.Warnings, error) {
+	return f.queryResultVal, nil, f.queryErr
+}
+
+func (f fakePrometheusClient) Alerts(ctx context.Context) (prometheusv1.AlertsResult, error) {
+	panic("implement me")
+}
+
+func (f fakePrometheusClient) AlertManagers(ctx context.Context) (prometheusv1.AlertManagersResult, error) {
+	panic("implement me")
+}
+
+func (f fakePrometheusClient) CleanTombstones(ctx context.Context) error {
+	panic("implement me")
+}
+
+func (f fakePrometheusClient) Config(ctx context.Context) (prometheusv1.ConfigResult, error) {
+	panic("implement me")
+}
+
+func (f fakePrometheusClient) DeleteSeries(ctx context.Context, matches []string, startTime time.Time, endTime time.Time) error {
+	panic("implement me")
+}
+
+func (f fakePrometheusClient) Flags(ctx context.Context) (prometheusv1.FlagsResult, error) {
+	panic("implement me")
+}
+
+func (f fakePrometheusClient) LabelNames(ctx context.Context, matches []string, startTime time.Time, endTime time.Time) ([]string, prometheusv1.Warnings, error) {
+	panic("implement me")
+}
+
+func (f fakePrometheusClient) LabelValues(ctx context.Context, label string, matches []string, startTime time.Time, endTime time.Time) (prometheusmodel.LabelValues, prometheusv1.Warnings, error) {
+	panic("implement me")
+}
+func (f fakePrometheusClient) QueryRange(ctx context.Context, query string, r prometheusv1.Range) (prometheusmodel.Value, prometheusv1.Warnings, error) {
+	panic("implement me")
+}
+
+func (f fakePrometheusClient) QueryExemplars(ctx context.Context, query string, startTime time.Time, endTime time.Time) ([]prometheusv1.ExemplarQueryResult, error) {
+	panic("implement me")
+}
+
+func (f fakePrometheusClient) Buildinfo(ctx context.Context) (prometheusv1.BuildinfoResult, error) {
+	panic("implement me")
+}
+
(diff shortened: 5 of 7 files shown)


pkg/operator/starter.go
```diff
@@ -52,6 +53,7 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 		operatorclient.OperatorNamespace,
 		operatorclient.TargetNamespace,
 		"kube-system",
+		"openshift-monitoring",
```
atiratree, line 56: should I add this as a constant to operatorclient ?

atiratree:
This PR triggers `GarbageCollectorObjectDeletionFailed` and degrades KCMo. Need to investigate

pkg/operator/gcwatchercontroller/prometheus_client.go
```diff
@@ -0,0 +1,75 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/transport"
+
+	"github.com/openshift/cluster-kube-controller-manager-operator/pkg/operator/operatorclient"
+)
+
+func newPrometheusClient(ctx context.Context, secretClient corev1client.SecretsGetter, configMapClient corev1client.ConfigMapsGetter) (prometheusv1.API, error) {
```
soltysh, line 24: I think that this isn't the right approach, it's probably ok for tests, but not for a production use. Would be nice to double check with @s-urbaniak about the right way to get the correct token.

soltysh (changes requested):

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
```diff
@@ -39,3 +39,27 @@ spec:
           for: 15m
           labels:
             severity: critical
+        - alert: GarbageCollectorSyncFailed
+          annotations:
+            summary: There was a problem with syncing the resources for garbage collection.
+            description: Garbage Collector had a problem with syncing and monitoring the available resources. Please see KubeControllerManager logs for more details.
+          expr: |
+            rate(garbagecollector_controller_garbagecollector_resources_sync_error_total{}[1h]) > 0
+          labels:
```
atiratree, line 48: I have removed the `for`s as they are not necessarily needed

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
```diff
@@ -39,3 +39,27 @@ spec:
...
+            severity: warning
+        - alert: GarbageCollectorObjectDeletionFailed
+          annotations:
+            summary: There was a problem with deleting an object.
+            description: Garbage Collector had some problems when attempting to delete an object. Please see KubeControllerManager logs for more details.
```
atiratree, line 53: is it okay to write see the logs as an advice?

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
```diff
@@ -39,3 +39,27 @@ spec:
...
+          expr: |
+            rate(workqueue_retries_total{name="garbage_collector_attempt_to_delete"}[5m]) > 0
+          labels:
+            severity: warning
+        - alert: GarbageCollectorObjectOrphaningFailed
+          annotations:
+            summary: There was a problem with orphaning an object.
+            description: Garbage Collector had a problem when attempting to orphan an object. Please see KubeControllerManager logs for more details.
+          expr: |
+            rate(workqueue_retries_total{name="garbage_collector_attempt_to_orphan"}[5m]) > 0
```
atiratree, line 63: I changed the configuration of these alerts (from `24h` to `5m`) but there are couple of problems with GC queues

1. It can have transitory errors, that are just for the purpose of retring the queue (https://github.com/kubernetes/kubernetes/blob/dd1a7893d4de47cf75cd6a11f38474b318dcc9b2/pkg/controller/garbagecollector/garbagecollector.go#L372). Since that can happen quite often we cannot observe the error rate change over long time. I have put 5 minutes there, so the alert will stop firing after that. Example errors from CI logs (they happen quite often in each run):


```
2022-05-11T21:42:35.109736421Z E0511 21:42:35.109708       1 garbagecollector.go:347] error syncing item &garbagecollector.node{identity:garbagecollector.objectReference{OwnerReference:v1.OwnerReference{APIVersion:"discovery.k8s.io/v1", Kind:"EndpointSlice", Name:"example-custom-endpoints-ff6nf", UID:"faae0457-1778-46d7-9cd1-6278ea6778f1", Controller:(*bool)(nil), BlockOwnerDeletion:(*bool)(nil)}, Namespace:"e2e-endpointslicemirroring-8071"}, dependentsLock:sync.RWMutex{w:sync.Mutex{state:0, sema:0x0}, writerSem:0x0, readerSem:0x0, readerCount:1, readerWait:0}, dependents:map[*garbagecollector.node]struct {}{}, deletingDependents:false, deletingDependentsLock:sync.RWMutex{w:sync.Mutex{state:0, sema:0x0}, writerSem:0x0, readerSem:0x0, readerCount:0, readerWait:0}, beingDeleted:false, beingDeletedLock:sync.RWMutex{w:sync.Mutex{state:0, sema:0x0}, writerSem:0x0, readerSem:0x0, readerCount:0, readerWait:0}, virtual:false, virtualLock:sync.RWMutex{w:sync.Mutex{state:0, sema:0x0}, writerSem:0x0, readerSem:0x0, readerCount:0, readerWait:0}, owners:[]v1.OwnerReference{v1.OwnerReference{APIVersion:"v1", Kind:"Endpoints", Name:"example-custom-endpoints", UID:"dc339dea-24f0-4a6f-a1f8-64b5d27e3ff6", Controller:(*bool)(0xc00a68b74c), BlockOwnerDeletion:(*bool)(0xc00a68b74d)}}}: endpointslices.discovery.k8s.io "example-custom-endpoints-ff6nf" not found
2022-05-11T21:42:35.116620452Z I0511 21:42:35.116595       1 garbagecollector.go:468] "Processing object" object="e2e-endpointslicemirroring-8071/example-custom-endpoints-ff6nf" objectUID=faae0457-1778-46d7-9cd1-6278ea6778f1 kind="EndpointSlice" virtual=false
2022-05-11T21:42:35.118599158Z I0511 21:42:35.116762       1 garbagecollector.go:580] "Deleting object" object="e2e-container-probe-3234/deployer-dockercfg-cqqnt" objectUID=94979b07-d97f-48c5-9a85-e5ac3fa71bdb kind="Secret" propagationPolicy=Background
2022-05-11T21:42:35.118599158Z E0511 21:42:35.116803       1 garbagecollector.go:347] error syncing item &garbagecollector.node{identity:garbagecollector.objectReference{OwnerReference:v1.OwnerReference{APIVersion:"v1", Kind:"Secret", Name:"builder-dockercfg-mm68z", UID:"7bf8d21e-79f5-4227-a0bd-bec4a113d826", Controller:(*bool)(nil), BlockOwnerDeletion:(*bool)(nil)}, Namespace:"e2e-container-probe-3234"}, dependentsLock:sync.RWMutex{w:sync.Mutex{state:0, sema:0x0}, writerSem:0x0, readerSem:0x0, readerCount:1, readerWait:0}, dependents:map[*garbagecollector.node]struct {}{}, deletingDependents:false, deletingDependentsLock:sync.RWMutex{w:sync.Mutex{state:0, sema:0x0}, writerSem:0x0, readerSem:0x0, readerCount:0, readerWait:0}, beingDeleted:false, beingDeletedLock:sync.RWMutex{w:sync.Mutex{state:0, sema:0x0}, writerSem:0x0, readerSem:0x0, readerCount:0, readerWait:0}, virtual:false, virtualLock:sync.RWMutex{w:sync.Mutex{state:0, sema:0x0}, writerSem:0x0, readerSem:0x0, readerCount:0, readerWait:0}, owners:[]v1.OwnerReference{v1.OwnerReference{APIVersion:"v1", Kind:"Secret", Name:"builder-token-x4t5h", UID:"3fe0d74b-cc1e-437f-8354-944c091a4646", Controller:(*bool)(0xc00966a2a8), BlockOwnerDeletion:(*bool)(0xc00966a2a9)}}}: secrets "builder-dockercfg-mm68z" not found
2022-05-11T21:42:35.123943615Z I0511 21:42:35.122021       1 garbagecollector.go:468] "Processing object" object="e2e-container-probe-3234/builder-dockercfg-mm68z" objectUID=7bf8d21e-79f5-4227-a0bd-bec4a113d826 kind="Secret" virtual=false2022-05-11T21:42:35.151610596Z E0511 21:42:35.151577       1 garbagecollector.go:347] error syncing item &garbagecollector.node{identity:garbagecollector.objectReference{OwnerReference:v1.OwnerReference{APIVersion:"v1", Kind:"Secret", Name:"deployer-dockercfg-cqqnt", UID:"94979b07-d97f-48c5-9a85-e5ac3fa71bdb", Controller:(*bool)(nil), BlockOwnerDeletion:(*bool)(nil)}, Namespace:"e2e-container-probe-3234"}, dependentsLock:sync.RWMutex{w:sync.Mutex{state:0, sema:0x0}, writerSem:0x0, readerSem:0x0, readerCount:1, readerWait:0}, dependents:map[*garbagecollector.node]struct {}{}, deletingDependents:false, deletingDependentsLock:sync.RWMutex{w:sync.Mutex{state:0, sema:0x0}, writerSem:0x0, readerSem:0x0, readerCount:0, readerWait:0}, beingDeleted:false, beingDeletedLock:sync.RWMutex{w:sync.Mutex{state:0, sema:0x0}, writerSem:0x0, readerSem:0x0, readerCount:0, readerWait:0}, virtual:false, virtualLock:sync.RWMutex{w:sync.Mutex{state:0, sema:0x0}, writerSem:0x0, readerSem:0x0, readerCount:0, readerWait:0}, owners:[]v1.OwnerReference{v1.OwnerReference{APIVersion:"v1", Kind:"Secret", Name:"deployer-token-mqt88", UID:"af5e1d10-e2be-4ae3-bf3a-f30c44e9bbb9", Controller:(*bool)(0xc00a711b28), BlockOwnerDeletion:(*bool)(0xc00a711b29)}}}: secrets "deployer-dockercfg-cqqnt" not found
2022-05-11T21:42:35.157711325Z I0511 21:42:35.156666       1 garbagecollector.go:468] "Processing object" object="e2e-container-probe-3234/deployer-dockercfg-cqqnt" objectUID=94979b07-d97f-48c5-9a85-e5ac3fa71bdb kind="Secret" virtual=false
```


2. The negative of the 5 minutes observation period is that the alert will miss long term errors since we have exponential backoff and no periodic resync in GC (as far as I could see).


So basically the usefullness of queue alerts is limited and could be useful only in a case something goes very wrong and we have a lot of new errors appearing very often (eg 1 a minute)

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
```diff
@@ -39,3 +39,30 @@ spec:
...
+            severity: warning
+        - alert: GarbageCollectorObjectOrphaningFailed
+          annotations:
+            summary: There was a problem with orphaning an object in the last 24h.
+            description: Garbage Collector had a problem when attempting to orphan an object in the last 24h. For more details, please see KubeControllerManagerOperator status for more details.
+          expr: |
+            rate(workqueue_retries_total{name="garbage_collector_attempt_to_orphan"}[24h]) > 0
+          for: 1m
+          labels:
+            severity: warning
```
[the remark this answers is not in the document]
atiratree: - so user can have better idea what part of GC is misbehaving
- please see comment above regarding the queues. Since they behave differently, I would prefer not to have them together with an alert for sync
- so we can have different observation periods: queues x sync

pkg/operator/gcwatchercontroller/prometheus_client.go
atiratree: I am using our serviceaccount token now, same as [etcd operator](https://github.com/openshift/cluster-etcd-operator/blob/c9977ae3bd788a9e9a595001c5bc24995f9d0175/pkg/operator/metriccontroller/client.go) does

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
atiratree: having before dismissed GC debug handler would help here a lot (https://github.com/kubernetes/kubernetes/pull/105705). We could easily distinguish between transient errors (simple queue requeues) and permanent errors by knowing the current state.

With current metrics
- we need a short observation period for the transient errors to quickly dismiss them (eg long alerts will botch CI)
- but we need long observation periods to guess the exponential backoff of the persistent errors

Which is hard to do without knowing the cause of each error. Even if we expose that somehow via some new metrics (which I think is impossilble without adding non trivial logic to gc) it is not possible to deal with exponential backoff easily - even if we can have 24h observation period it is not the best solution as the alert will stop if nobody gives it attention for some time

These limitation were not clear to me when we discussed it last time.

The other alternative is just checking the dependencies of GC like APIServices, as we also discussed. But I am not favoring this approach since we would not have a real insight into what GC is doing wrt errors.

@soltysh thoughts?

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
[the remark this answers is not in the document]
soltysh: sgtm

pkg/operator/gcwatchercontroller/prometheus_client.go
soltysh: Yeah, this approach is definitely better.

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
soltysh: @atiratree let's leave that APIServices as a last resort, for now is it possible to pair this metric with some others related with workqueue, `workqueue_depth`, for example. In that case transient errors will grow the depth, but will eventually stop, whereas in the long observation both can potentially grow. Or maybe a different combination, worth syncing with @dgrisonnet about that it.

ravisantoshgudimetla: I think for the alert to be at warning level, we decided to use 60 min as timeline.

https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/alerting-consistency.md#warning-alerts

> Even if we expose that somehow via some new metrics (which I think is impossilble without adding non trivial logic to gc) it is not possible to deal with exponential backoff easily

Would it be easier to split the metric and make it much more specialized to indicate that the discovery of objects is failing(perhaps a failing webhook conversion) by having a metric [here](https://github.com/atiratree/kubernetes/blob/2fc401d0a295dd30ec73e99dbcdbba6a08268d6e/pkg/controller/garbagecollector/garbagecollector.go#L763)

atiratree: here the period has a bit different meaning, since we use it just for stopping caring about the error - and we for sure know that the for example sync error has happened. 

> Would it be easier to split the metric and make it much more specialized to indicate that the discovery of objects is failing(perhaps a failing webhook conversion) by having a metric

I am not sure if another metric in this place would be beneficial yet

atiratree: I went over all available metrics and there seems not to be a solution:

`workqueue_retries_total`: does not work due to the already mentioned reasons
`workqueue_depth` and `workqueue_queue_duration_seconds`: does not work - since retried items are not really in the queue and thus are not registred via the metric and are just waiting to be added to the queue after a backoff
`workqueue_work_duration_seconds`: not usable - tracks only work duration
`workqueue_adds_total`: not usable - tracks only adds
`workqueue_unfinished_work_seconds`: not usable (usually 0)
`workqueue_longest_running_processor_seconds`: not usable (usually 0)

So wrt alternatives:

It cannot be simply solved just by looking at APIServices to detect they are down. We might still have old items in GC (errors / retries) even after the APIServices get removed,


To reiterate, the problem is that the same rest mapping errors can be both transient and persistent (https://github.com/kubernetes/kubernetes/blob/dd1a7893d4de47cf75cd6a11f38474b318dcc9b2/pkg/controller/garbagecollector/garbagecollector.go#L372-L374) as this depends on an outside actor/sync.
I think it could be solved generically by a mechanism that would track all items that are waiting for a requeue/retry and report a metric if they overcome an arbitrary threshold (eg. 30s) to inhibit counting of legitimate retries.

This would require some changes in a GC to wire this. @soltysh @ravisantoshgudimetla  thoughts?

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
dgrisonnet: just to make sure this is what you want:
```
rate(garbagecollector_controller_garbagecollector_resources_sync_error_total{}[1h]) > 0
```
means that if any sync error occur, the alert will fire right away
whereas:
```
rate(garbagecollector_controller_garbagecollector_resources_sync_error_total{}[5m]) > 0

for: 1h
```
means that if at least one sync error occurs every 5 minutes during the course of an hour, the alert will fire

The first approach seems very noisy but maybe it is intended.

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
dgrisonnet: Yes, it is.

The best is always to write a dedicated runbook to provide as much information as possible, but having a comprehensive description is a good start.

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
```diff
@@ -39,3 +39,27 @@ spec:
...
+          expr: |
+            rate(garbagecollector_controller_garbagecollector_resources_sync_error_total{}[1h]) > 0
+          labels:
+            severity: warning
+        - alert: GarbageCollectorObjectDeletionFailed
+          annotations:
+            summary: There was a problem with deleting an object.
+            description: Garbage Collector had some problems when attempting to delete an object. Please see KubeControllerManager logs for more details.
+          expr: |
+            rate(workqueue_retries_total{name="garbage_collector_attempt_to_delete"}[5m]) > 0
```
dgrisonnet, line 55: This might be very noisy. The alert will fire whenever a retry is detected. I am not sure how common it is for the GC, but in some other contexts, it is very normal to retry. Having a `for` clause here might make the alert more resilient to problems that are being resolved right away.

In general, warning level alerts, tend to have less priority for cluster admins, so waiting for even an hour before looking into one is pretty normal so as developers writing alerts we should thrive to reduce the noise for them and eliminate any kind of false positive.

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
dgrisonnet: I am not sure I have a complete understanding of the problem you are trying to solve, but I am happy to discuss it offline if you want.

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
atiratree: Yes it is insufficient and manifests problems you are describing. I have been trying to tweak these two (`garbage_collector_attempt_to_delete` and `garbage_collector_attempt_to_delete` ) and see how they perform in CI.

Ultimately I found out we probably have to bring in a different solution instead of this one: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/623#discussion_r879920985

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
atiratree: I have put `for` period back to `60m` after a discussion with @dgrisonnet since it is not that important to fire an alert right away since non working gc is not that critical. We can report the alert after getting persistent failures for one hour

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
atiratree: After discussing this with @ravisantoshgudimetla and @dgrisonnet, the best course is to introduce new metrics in GC for the delete and orphan queue (probably a histogram for the backoffs).

pkg/operator/gcwatchercontroller/prometheus_client.go
```diff
@@ -0,0 +1,60 @@
...
+	routerCAConfigMap, err := configMapClient.ConfigMaps(operatorclient.GlobalMachineSpecifiedConfigNamespace).Get(ctx, "service-ca", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	bundlePEM := []byte(routerCAConfigMap.Data["ca-bundle.crt"])
+
+	roots := x509.NewCertPool()
+	roots.AppendCertsFromPEM(bundlePEM)
+
+	client, err := prometheusapi.NewClient(prometheusapi.Config{
```
soltysh, line 40: Nit: you're swallowing the error from here.

atiratree: yikes! Thanks for catching it

soltysh (approved):
/lgtm

soltysh (approved):
/lgtm

PR #627, opened 2022-05-16
e2e: Fix test pod disruption budget at limit alert

atiratree:
in 5% [cases](https://search.ci.openshift.org/?search=.*TestPodDisruptionBudgetAtLimitAlert.*&maxAge=336h&context=1&type=bug%2Bjunit&name=e2e-aws-operator&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job) there is a SA missing in the test namespace:

```
 operator_test.go:103: pods "pdbtest-9t6n7" is forbidden: error looking up service account pdbtest-9t6n7/default: serviceaccount "default" not found
```

I could not reproduce this at my cluster, but this should fix it.

### DIFF
--- a/test/library/namespace.go
+++ b/test/library/namespace.go
@@ -1,0 +1,59 @@
+package library
+
+import (
+	"context"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


test/library/namespace.go
```diff
@@ -0,0 +1,59 @@
+package library
```
ravisantoshgudimetla, line 1: good call on making this exportable package to be used by e2e and e2e-preferred-host. We can move client creation logic here.

ravisantoshgudimetla (approved):
/lgtm
/approve

PR #629, opened 2022-05-25
Bug 2087684: Reject transition from/to extreme latency profiles (default<->low)

swghosh:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2087684

- added another suppress func that rejects extreme profile transition attempts 
  - Default -> LowUpdateSlowReaction
  - LowUpdateSlowReaction -> Default

This is in favour of cluster stability as mentioned in EP [1][2].

[1] https://github.com/openshift/enhancements/blob/master/enhancements/worker-latency-profile/worker-latency-profile.md#default---lowupdateslowreaction
[2] https://github.com/openshift/enhancements/blob/master/enhancements/worker-latency-profile/worker-latency-profile.md#lowupdateslowreaction---default

Signed-off-by: Swarup Ghosh <swghosh@redhat.com>


pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -110,11 +120,14 @@ func NewConfigObserver(
...
 			nodeobserver.NewLatencyProfileObserver(
 				node.LatencyConfigs,
-				nodeobserver.NewSuppressConfigUpdateUntilSameProfileFunc(
-					operatorClient.(v1helpers.StaticPodOperatorClient),
-					kubeInformersForNamespaces.ConfigMapLister().ConfigMaps(operatorclient.TargetNamespace),
-					node.LatencyConfigs,
-				),
+				[]nodeobserver.ShouldSuppressConfigUpdatesFunc{
+					extremeProfileSuppressor,
+					nodeobserver.NewSuppressConfigUpdateUntilSameProfileFunc(
```
atiratree, line 125: please make a variable for this as well

swghosh: Added.

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -110,11 +126,10 @@ func NewConfigObserver(
 			network.ObserveServiceClusterIPRanges,
 			nodeobserver.NewLatencyProfileObserver(
 				node.LatencyConfigs,
-				nodeobserver.NewSuppressConfigUpdateUntilSameProfileFunc(
-					operatorClient.(v1helpers.StaticPodOperatorClient),
-					kubeInformersForNamespaces.ConfigMapLister().ConfigMaps(operatorclient.TargetNamespace),
-					node.LatencyConfigs,
-				),
+				[]nodeobserver.ShouldSuppressConfigUpdatesFunc{
+					extremeProfileSuppressor,
```
atiratree, line 130: let's put a comment here that the first one should be the more important one as it will resolve first. @swghosh can you also please put such a comment to KASo?

swghosh: Ahh, for KASO it won't be required as KASO don't require `extremeProfileSuppressor` and it'll just have the `differentConfigProfileSuppressor`. Because for KubeAPIServer: Medium and Low has the same set of arg val pairs, thus no rejection is required there!

[1] https://coreos.slack.com/archives/C0392N52Q5P/p1653567422152599?thread_ts=1653567121.768759&cid=C0392N52Q5P
[2] https://github.com/openshift/enhancements/blob/master/enhancements/worker-latency-profile/worker-latency-profile.md#mediumupdateaveragereaction---lowupdateslowreaction

atiratree: Ok cool, thanks for the comment

PR #631, opened 2022-06-01
Bug 1902307: Let vsphere-legacy-cloud-provider update node objects

lobziik:
Starting kube 1.24, legacy vsphere cloud-provider requires
permission to update nodes in order to set topology related labels.

For more info see https://github.com/kubernetes/kubernetes/pull/101028


pkg/operator/starter.go
```diff
@@ -104,6 +104,9 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 			"assets/kube-controller-manager/csr_approver_clusterrolebinding.yaml",
 			"assets/kube-controller-manager/gce/cloud-provider-role.yaml",
 			"assets/kube-controller-manager/gce/cloud-provider-binding.yaml",
+			"assets/kube-controller-manager/vsphere/legacy-cloud-provider-sa.yaml",
```
ingvagabund, line 107: In the GCE case the additional RBAC includes events and service status. Hard something that might bring down a cluster. Though extending the RBACs with node's update/patch might cause trouble. Would it make more sense to deploy the additional RBACs only on vsphere?

lobziik: Done in https://github.com/openshift/cluster-kube-controller-manager-operator/pull/631/commits/575ddc797fdb817c6bae47d91526127f97d170b0

JoelSpeed (commented):
/lgtm

soltysh (approved):
/approve

PR #633, opened 2022-06-23
detect ObjectDeletion and ObjectOrphaning errors by Alerts and  by GC watcher

atiratree:
This is dependendent on upstream changes in https://github.com/kubernetes/kubernetes/pull/110329

Original adjacent discussion for these new metrics: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/623#discussion_r871633087


atiratree:
we are not pursuing this anymore in favor of other more generic approaches

PR #634, opened 2022-06-23
add PodsUnableToDelete alert

atiratree:


manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
```diff
@@ -39,6 +39,14 @@ spec:
           for: 15m
           labels:
             severity: critical
+        - alert: PodsUnableToDelete
+          annotations:
+            summary: Some pods are unable to be deleted.
+            description: Some pods have been unable to be deleted for last 24h (since their grace period for termination has been reached).
+          expr: |
+            count by (namespace) (time() - kube_pod_deletion_timestamp > 86400)
```
atiratree, line 47: is having this for every pod useful? Maybe aggregation over namespaces would be better in case there is a large number of pods are stuck?

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
```diff
@@ -39,6 +39,14 @@ spec:
...
+          labels:
+            severity: warning
```
atiratree, line 49: is `for:` useful here? since we are just waiting when the expression reaches over 24h.

wking: Right, we don't need `for`, because we're handling the "yeah, this is not going away" inertia in the `expr`.

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
```diff
@@ -39,6 +39,14 @@ spec:
           for: 15m
           labels:
             severity: critical
+        - alert: PodsUnableToDelete
+          annotations:
+            summary: Some pods are unable to be deleted.
+            description: Some pods have been unable to be deleted for last 24h (since their grace period for termination has been reached).
```
wking, line 45: We don't actually have access to their termination grace period, do we?  These are pods that were initially asked to delete 24h ago.  I expect in most cases, the termination grace period will be smaller than 24h.  But maybe not?  If we can find a way to start the timer when they hit their termination grace period, that would be great.

atiratree: actually the grace period is embedded in the deletionTimestamp, so basically the deletion timestamp is in the future. And after that there is probably something wrong going on. So I guess we could get with even much smaller window than 24h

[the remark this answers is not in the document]
atiratree: I think it would be best to introduce a runbook_url for this and keep it general here. For finding the pods, maybe we can just point them to the query without the `namespace` aggregation?

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
dgrisonnet: Another approach could be to just alert whenever there is at least one pod stuck for more than 24h and provide a query in a runbook to get the list of all the pods that are stuck.

Alert expression would be:
```
time() - min(kube_pod_deletion_timestamp) > 86400)
```

Runbook expression would be:
```
time() - group by (namespace, pod, instance) (kube_pod_deletion_timestamp) > 86400
```

PR #636, opened 2022-07-06
Bug 1968555: Decouple KCM-O's status from monitoring stack

ravisantoshgudimetla:
https://github.com/openshift/cluster-kube-controller-manager-operator/pull/623/files helped in making GC debugging clear but it introduced dependency on monitoring stack. In order to not have that dependency I am thinking of not updating status of KCM-O when monitoring stack is disabled or not needed.

cc @atiratree @soltysh


soltysh (approved):
/lgtm
/approve
/bugzilla refresh

PR #639, opened 2022-07-13
Make KCM-O conditionally dependent on monitoring stack availability

ravisantoshgudimetla:
https://github.com/openshift/cluster-kube-controller-manager-operator/pull/636/files completely removed the dependency on monitoring to unblock the OpenShift locals team. However, it is a functionality that will help uncover a lot of bugs in normal OpenShift cluster deployments and we wanted to have it back after we discuss it with the monitoring team. This PR introduces a conditional dependency on the monitoring stack and if the monitoring is not available instead of throwing an error, we log and return immediately.

cc @atiratree @soltysh @praveenkumar @deads2k


praveenkumar:
@ravisantoshgudimetla I will test this with Openshift Local once this PR land to one of the 4.12 nightly and let you know. Or I can create the payload with this PR and try (which is time consuming), whatever you prefer.

pkg/operator/gcwatchercontroller/gcwatcher_controller.go
```diff
@@ -102,16 +111,24 @@ func (c *GarbageCollectorWatcherController) syncWorker(ctx context.Context, sync
...
 
-	prometheusClient, err := newPrometheusClient(ctx, c.configMapClient)
-	if err != nil && errors.IsNotFound(err) {
-		// Prometheus client when failed to instantiate should not result in error being generated. This makes sure we don't
-		// generate excessive events.
-		return nil
-	} else if err != nil {
-		return err
+	if c.prometheusClient == nil {
+		// Don't create prometheus client on every sync, re-use from struct, this will also
```
atiratree, line 115: we have to instantiate the client every time to detect the CA change

alternatively we could detect that the config map changed and reinitialize the client only in that case

pkg/operator/gcwatchercontroller/gcwatcher_controller.go
```diff
@@ -183,7 +200,7 @@ func checkFiringAlerts(ctx context.Context, requiredAlertsSet sets.String, prome
 		klog.Warningf("received warnings when querying alerts: %v\n", strings.Join(warnings, ", "))
 	}
 	if err != nil {
-		return fmt.Errorf("error querying alerts: %v", err)
+		return err
```
atiratree, line 203: why reverting a more verbose message on what is wrong?

pkg/operator/gcwatchercontroller/gcwatcher_controller_test.go
```diff
@@ -243,3 +254,98 @@ func TestCheckFiringAlerts(t *testing.T) {
...
+			}
+			listers := configlisters.NewClusterOperatorLister(indexer)
+			tc.gc.clusterLister = listers
+			eventRecorder := events.NewInMemoryRecorder("dummy")
+			syncContext := factory.NewSyncContext(controllerName, eventRecorder)
+			syncContext.Queue().Add(invalidateAlertingRulesCacheKey)
+			err := tc.gc.sync(context.TODO(), syncContext)
+			if tc.expectErr && err == nil {
+				t.Fatal("expected error but did not get one")
+			} else if tc.expectErr && errors.Is(err, tc.expectedErrorMsg) {
```
atiratree, line 338:
```suggestion
			} else if tc.expectErr && !errors.Is(err, tc.expectedErrorMsg) {
```

shouldn't it be more something like this?

atiratree: also might be nicer to wrap this whole in `if tc.expectErr` and do the else branch as well

pkg/operator/gcwatchercontroller/gcwatcher_controller.go
ravisantoshgudimetla: `prometheusClient.Query()` is already adding that context.

So I removed it.

pkg/operator/gcwatchercontroller/gcwatcher_controller.go
```diff
@@ -102,16 +122,24 @@ func (c *GarbageCollectorWatcherController) syncWorker(ctx context.Context, sync
...
 
-	prometheusClient, err := newPrometheusClient(ctx, c.configMapClient)
-	if err != nil && errors.IsNotFound(err) {
-		// Prometheus client when failed to instantiate should not result in error being generated. This makes sure we don't
-		// generate excessive events.
-		return nil
-	} else if err != nil {
-		return err
+	if !c.promConnectivity.useCachedClient {
+		// Don't create prometheus client on every sync unless explicitly asked, re-use from struct, this will also
```
atiratree, line 126: we can switch this comment to: `use cached only for testing` or soemthing like that

pkg/operator/gcwatchercontroller/gcwatcher_controller.go
```diff
@@ -102,16 +122,24 @@ func (c *GarbageCollectorWatcherController) syncWorker(ctx context.Context, sync
...
+		// help in unit testing.
+		prometheusClient, err := newPrometheusClient(ctx, c.configMapClient)
+		if err != nil {
+			// Prometheus client when failed to instantiate should not result in error being generated. We can reach
+			// this stage if CMO is disabled day-2  and thanos services are removed after cluster installation
+			// has happened.
+			// TODO: In future, cluster operators can have status which states if they are managed by CVO or not
+			//		and we can use to represent failure.
+			klog.Infof("failed to instantiate prometheus client. Thanos is not queriable at the moment with %v",
```
atiratree, line 135: better to put warn or error here

pkg/operator/gcwatchercontroller/gcwatcher_controller_test.go
```diff
@@ -243,3 +254,102 @@ func TestCheckFiringAlerts(t *testing.T) {
...
+			syncContext := factory.NewSyncContext(controllerName, eventRecorder)
+			syncContext.Queue().Add(invalidateAlertingRulesCacheKey)
+			err := tc.gc.sync(context.TODO(), syncContext)
+			if tc.expectErr {
+				if err == nil {
+					t.Fatal("expected error but did not get one")
+				} else if tc.expectedErrorMsg.Error() != err.Error() {
+					t.Fatalf("expected err %v but got %v", tc.expectedErrorMsg, err)
+				}
+			}
```
atiratree, line 344: let's check if we got an error in the else branch

ravisantoshgudimetla: I think this is what you wanted?

pkg/operator/gcwatchercontroller/gcwatcher_controller.go
ravisantoshgudimetla: I added a comment during the struct declaration but I added one more here. Let me know what you think of this comment.

pkg/operator/gcwatchercontroller/gcwatcher_controller_test.go
atiratree: +1

pkg/operator/gcwatchercontroller/gcwatcher_controller.go
atiratree: looks good, although it might be better to put it before the if, since it is negated

PR #642, opened 2022-07-27
Reset Go mod after openstack library-go bump

JoelSpeed:
Following on from #641, we force merged dependencies from a branch. This resets back to the updated library-go commit and removes the replace statement.

No diff to the vendor.


atiratree:
ci fix #643  was merged
/retest

PR #645, opened 2022-08-02
[release-4.10] Bug 2114587: add runbook urls to KCM-o alerts

atiratree:
cherrypick of #644


ingvagabund:
Given https://github.com/openshift/cluster-kube-controller-manager-operator/pull/646 is going in, we can merge this no-op change since a new KCM respin will be performed anyway.
/approve
/label backport-risk-assessed

PR #650, opened 2022-08-24
Bug 2118286: always report and reconcile GarbageCollectorDegraded condition

atiratree:


pkg/operator/gcwatchercontroller/gcwatcher_controller.go
```diff
@@ -87,21 +87,24 @@ func (c *GarbageCollectorWatcherController) sync(ctx context.Context, syncCtx fa
...
+		Reason: "AsExpected",
+	}
 	_, err := c.clusterLister.Get("monitoring")
 	if err != nil && errors.IsNotFound(err) {
 		klog.V(5).Info("Monitoring is disabled in the cluster and a diagnostic of the garbage collector is not working. Please look at the kcm logs for more information to debug the garbage collector further")
-		return nil
+		// Disabled monitoring works as expected and is not degraded
+		condition.Reason = "MonitoringDisabled"
+		_, _, updateErr := v1helpers.UpdateStatus(ctx, c.operatorClient, v1helpers.UpdateConditionFn(condition))
+		return updateErr
```
ingvagabund, line 101: Normally this does not have any effect on clusters where the monitoring stack is either absent from the beginning or always present from the beginning. This covers the cases where the monitoring stack is present but then removed. So once removed the `syncWorker` method starts producing errors which are acceptable but with the current code the degraded condition is set to true instead of false.

atiratree: the only change on the clusters without monitoring will be that it will have `GarbageCollectorDegraded` with always set to false. I think this should not be an issue and at least the conditions will be more consistent acrross environments

ingvagabund: That's acceptable.

PR #652, opened 2022-08-30
k8s 1.25.0

soltysh:


soltysh:
/test e2e-aws-operator
the panic in ocmo is fixed in https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/259

PR #602, opened 2022-02-09
Update to use configmapleases

ravisantoshgudimetla:
Update to use configmapleases instead of leases

PR #661, opened 2022-11-04
OCPBUGS-3283: remove unnecessary RBAC

atiratree:
- was used for making a switch from configMaps to leases in leader election

we noticed this in https://coreos.slack.com/archives/CC3CZCQHM/p1667571136730989
Also you can see the TODOs in the removed files (that are still present for the controller to work).

- https://github.com/openshift/cluster-kube-controller-manager-operator/blob/08d7aed96479b280a23c1e7c2075c98164fc95bd/bindata/assets/kube-controller-manager/leader-election-kube-controller-manager-role-kube-system.yaml#L5
- https://github.com/openshift/cluster-kube-controller-manager-operator/blob/08d7aed96479b280a23c1e7c2075c98164fc95bd/bindata/assets/kube-controller-manager/leader-election-kube-controller-manager-rolebinding-kube-system.yaml#L5

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -97,8 +97,6 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 			"assets/kube-controller-manager/leader-election-rolebinding.yaml",
 			"assets/kube-controller-manager/leader-election-cluster-policy-controller-role.yaml",
 			"assets/kube-controller-manager/leader-election-cluster-policy-controller-rolebinding.yaml",
-			"assets/kube-controller-manager/leader-election-kube-controller-manager-role-kube-system.yaml",
-			"assets/kube-controller-manager/leader-election-kube-controller-manager-rolebinding-kube-system.yaml",
 			"assets/kube-controller-manager/namespace-security-allocation-controller-clusterrole.yaml",
 			"assets/kube-controller-manager/namespace-security-allocation-controller-clusterrolebinding.yaml",
 			"assets/kube-controller-manager/podsecurity-admission-label-syncer-controller-clusterrole.yaml",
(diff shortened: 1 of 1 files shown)


pkg/operator/starter.go
```diff
@@ -118,6 +114,19 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 		(&resourceapply.ClientHolder{}).WithKubernetes(kubeClient),
 		operatorClient,
 		cc.EventRecorder,
+	).WithConditionalResources(
+		bindata.Asset,
+		[]string{
+			// TODO: remove all of these leader-election entries and files in 4.13
```
atiratree, line 120: would need to change the TODO to 4.14 in case we decide not to backport it

ingvagabund: We switched to leases in 4.11 (https://github.com/openshift/cluster-kube-controller-manager-operator/pull/602). So the create rbac for configmaps is not needed starting that version. Though, I would not temper with it in 4.11 unless there's a high priority request for it. Backporting to 4.12 is fine.

ingvagabund: Nevertheless, this removes RBAC for the leases for CPC as well. Can you more elaborate (preferable in the PR description) why the RBACs are no longer needed? If possible, linking the code changes which proves the RBACs are no longer needed.

atiratree: regarding the backport, yes I was planning only to 4.12 branch - so it will have a chance to remove the RBAC in 4.12 and we remove it in 4.13

Good idea mentioning the TODOs in the description. I also wanted to link the slack thread but forgot ^^. Updated.

atiratree:
/hold
issue with seemingly not relevant CI failures

debugging in 
- https://github.com/openshift/cluster-kube-controller-manager-operator/pull/668
- https://github.com/openshift/cluster-kube-controller-manager-operator/pull/669

atiratree:
@ingvagabund I have found that this not the case for cluster-policy-controller since this is the only RBAC that gives configmaps/leases permissions to it.

I left only the KCM changes in this PR and keeping the CPC RBAC intacted

Nevertheless there were too wide permissions in CPC so I decided to fix that in a separate PR https://github.com/openshift/cluster-kube-controller-manager-operator/pull/670

PR #662, opened 2022-11-04
Remove stale conditions controller

atiratree:


pkg/operator/starter.go
```diff
@@ -231,16 +230,6 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 	}
 	saTokenController := certrotationcontroller.NewSATokenSignerController(operatorClient, kubeInformersForNamespaces, kubeClient, cc.EventRecorder)
 
-	staleConditions := staleconditions.NewRemoveStaleConditionsController(
```
ingvagabund, line 234: There are other places where `NewRemoveStaleConditionsController` is still used: https://github.com/search?q=org%3Aopenshift+NewRemoveStaleConditionsController&type=code. Would you please check whether it can be removed from other components as well?

ingvagabund: @deads2k is there anything else left to do or are we safe to drop these?

atiratree: not sure about all of the repos from the query, but I suppose I could start with the ours where the comments are clear

PR #663, opened 2022-11-09
OCPBUGS-3985: Enforce PSA when techpreview is enabled

deads2k:
This is needed to trigger the cluster-policy-controller changes.

relies on https://github.com/openshift/api/pull/1341


go.mod
```diff
@@ -115,3 +115,7 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
+
+replace github.com/openshift/api => github.com/deads2k/api v0.0.0-20221109205221-08485b73dfa0
+
+replace github.com/openshift/library-go => github.com/deads2k/library-go v0.0.0-20221109210222-13d8674239e1
```
stlaz, line 121: are both of these merged now?

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -122,6 +122,11 @@ func NewConfigObserver(
 				openShiftOnlyFeatureGates,
 				[]string{"extendedArguments", "feature-gates"},
 			),
+			featuregates.NewObserveFeatureFlagsFunc(
```
stlaz, line 125: Will this be pushing OpenShift-specific feature-gates to KCM? How's it going to influence the feature gates pushed from the above feature flags observer?

I think we need at least a comment explaining why there are two feature flags observers

deads2k: Comment added and content updated.

PR #664, opened 2022-11-21
DEBUG: sync openshift/library-go

ingvagabund:
To test changes in https://github.com/openshift/library-go/pull/1421.


pkg/cmd/render/render.go
```diff
@@ -246,7 +246,7 @@ func (r *renderOpts) Run() error {
 	}
 	extendedArguments := targetconfigcontroller.GetKubeControllerManagerArgs(kubeControllerManagerConfig)
 	for _, arg := range extendedArguments {
-		renderConfig.ExtendedArguments += fmt.Sprintf("    - %s\n", arg)
+		renderConfig.ExtendedArguments += fmt.Sprintf("\n    - %s", arg)
```
atiratree, line 249: tested it and it seems we do not need the final newline

PR #678, opened 2022-12-19
OCPBUGS-5006: add leader-elect-renew-deadline into defaultconfit.yaml

Rupesh-git-eng:
The default timeout 5s is not enough to sustain in case of primary DNS failure.
https://issues.redhat.com/browse/OCPBUGS-5006

### DIFF
--- a/bindata/assets/config/defaultconfig.yaml
+++ b/bindata/assets/config/defaultconfig.yaml
@@ -19,6 +19,8 @@ extendedArguments:
   - "true"
   leader-elect-retry-period:
   - "3s"
+  leader-elect-renew-deadline:
+  - "12s"
   leader-elect-resource-lock:
   - "configmapsleases"
   controllers:


bindata/assets/config/defaultconfig.yaml
```diff
@@ -19,6 +19,8 @@ extendedArguments:
   - "true"
   leader-elect-retry-period:
   - "3s"
+  leader-elect-renew-deadline:
+  - "12s"
```
atiratree, line 23: @Rupesh-git-eng can you please comment a short disclaimer here why we do this? Eg. This will make the request timeout 6s and is needed to have a higher value than a 5s DNS timeout.

Or something similar in that sense?

atiratree:
you need to update the unit tests as well

atiratree:
/retitle OCPBUGS-5006: add leader-elect-renew-deadline into defaultconfit.yaml

bindata/assets/config/defaultconfig.yaml
Rupesh-git-eng: This will increase the default timeout from 5s to 6s and that extra one second is sufficient to wait for an answer from the secondary DNS server if the primary DNS server is failed in /etc/resolv.conf of the master node. 

"https://api-int.xxxx.xxx.com:6443/api/v1/namespaces/kube-system/configmaps/kube-controller-manager?timeout=5s"": context deadline exceeded

Rupesh-git-eng:
> you need to update the unit tests as well

atiratree:
@Rupesh-git-eng to fix the unit tests it should be enough to just add the flag and value to the render test https://github.com/openshift/cluster-kube-controller-manager-operator/blob/9cf88f00d9310b431561485135c42da712435fd0/pkg/cmd/render/render_test.go#L161

Rupesh-git-eng:
> @Rupesh-git-eng to fix the unit tests it should be enough to just add the flag and value to the render test
> 
> https://github.com/openshift/cluster-kube-controller-manager-operator/blob/9cf88f00d9310b431561485135c42da712435fd0/pkg/cmd/render/render_test.go#L161

Noted, will check ASAP.

bindata/assets/config/defaultconfig.yaml
```diff
@@ -19,6 +19,8 @@ extendedArguments:
   - "true"
   leader-elect-retry-period:
   - "3s"
+  leader-elect-renew-deadline:
+  - "12s" # This will increase timeout value from 5s to 6s that is required in case primary dns server in resolv.conf fail.
```
atiratree, line 23: thanks for adding the message. I think it would rather make sense to either say `from default 5s` or remove `from 5s` entirely

atiratree:
e2e-aws-ovn-ccm-install failure suggest a flake

/approve

PR #685, opened 2023-01-05
do not clobber bootstrap kcm log on restart

sanchezl:

PR #686, opened 2023-01-06
honor feature gates during bootstrapping

deads2k:
like https://github.com/openshift/cluster-kube-apiserver-operator/pull/1419 for kcm.  this one is also on the path to PSA enforcement via featuregate.


pkg/cmd/render/render.go
```diff
@@ -3,6 +3,7 @@ package render
 import (
 	"errors"
 	"fmt"
+	configv1 "github.com/openshift/api/config/v1"
```
stlaz, line 6: move outside stdlib imports

deads2k:
merge after https://github.com/openshift/cluster-kube-controller-manager-operator/pull/685

deads2k:
tagging per slack, holding so the kube-update blocking PR can merge first.

PR #698, opened 2023-02-14
Try to query rules before returning the prometheus client

eranco74:
During the cluster installation kube-controller-manager cluster operator is degraded for a few minutes due to:
```
GarbageCollectorDegraded: error fetching rules: Get "https://thanos-querier.openshift-monitoring.svc:9091/api/v1/rules": dial tcp 172.30.119.108:9091: connect: connection refused
```
The GarbageCollectorDegraded status is set to true because the prometheusClient is successfully initialized, but gets a connection refused once we try to query the rules.
Note that as long the client initialization fails the kube-controller-manger won't set the  GarbageCollectorDegraded to true.

### DIFF
--- a/pkg/operator/gcwatchercontroller/prometheus_client.go
+++ b/pkg/operator/gcwatchercontroller/prometheus_client.go
@@ -62,6 +62,7 @@ func newPrometheusClient(ctx context.Context, configMapClient corev1client.Confi
 	if err != nil {
 		return nil, nil, err
 	}
-
-	return prometheusv1.NewAPI(client), t, nil
+	prometheusClient := prometheusv1.NewAPI(client)
+	_, err = prometheusClient.Rules(ctx)
+	return prometheusClient, t, err
 }


pkg/operator/gcwatchercontroller/prometheus_client.go
```diff
@@ -62,6 +62,9 @@ func newPrometheusClient(ctx context.Context, configMapClient corev1client.Confi
 	if err != nil {
 		return nil, nil, err
 	}
-
-	return prometheusv1.NewAPI(client), t, nil
+	prometheusClient := prometheusv1.NewAPI(client)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
tsorya, line 66: Maybe better to use already provided context?

pkg/operator/gcwatchercontroller/prometheus_client.go
```diff
@@ -62,6 +62,7 @@ func newPrometheusClient(ctx context.Context, configMapClient corev1client.Confi
 	if err != nil {
 		return nil, nil, err
 	}
-
-	return prometheusv1.NewAPI(client), t, nil
+	prometheusClient := prometheusv1.NewAPI(client)
+	_, err = prometheusClient.Rules(ctx)
+	return prometheusClient, t, err
```
atiratree, line 67: this would introduce false positives when thanos/prometheus is down for any reason. In that scenario we could not infer the status of garbage collector and the condition would incorrectly be set to false

eranco74: At first, I thought that this is exactly what [this error](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/7f96ac5fb523366079e49e838ae4df83d31709d6/pkg/operator/gcwatchercontroller/gcwatcher_controller.go#L175)  handles, but now I notice that I miss interpret the comment above it.

eranco74: Any  way we can get the GarbageCollectorWatcherController sync method called more often until prometheusClient can actually query the rules?
Otherwise, the kube-controller-manager operator isn't available due to  GarbageCollectorDegraded condition for few minutes.

atiratree: @eranco74 I am trying an alternative approach that should not degrade KCM at all when monitoring stack rollout is in progress: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/706

PR #706, opened 2023-03-07
OCPBUGS-7440: do not degrade KCM when when monitoring stack rollout is in progress

atiratree:
because we have a resync period of 5m minutes it might take this much time before the gcwatcher controller successfully connects to the prometheus API and drops KCM degraded state. This can potentially slow down the cluster installation/upgrade as shown in the jira issue.

this is an alternative solution to https://github.com/openshift/cluster-kube-controller-manager-operator/pull/698

### DIFF
--- a/pkg/operator/gcwatchercontroller/gcwatcher_controller.go
+++ b/pkg/operator/gcwatchercontroller/gcwatcher_controller.go
@@ -3,6 +3,7 @@ package gcwatchercontroller
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strings"
 	"sync"
 	"time"
@@ -18,9 +19,11 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 
+	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	configinformers "github.com/openshift/client-go/config/informers/externalversions"
 	configlisters "github.com/openshift/client-go/config/listers/config/v1"
+	clusteroperatorhelpers "github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
@@ -52,6 +55,7 @@ const (
 	controllerName                  = "garbage-collector-watcher-controller"
 	invalidateAlertingRulesCacheKey = "__internal/invalidateAlertingRulesCacheKey"
 	invalidateAlertingRulesPeriod   = 12 * time.Hour
+	monitoringStackDeployTimeout    = time.Hour
 )
 
 func NewGarbageCollectorWatcherController(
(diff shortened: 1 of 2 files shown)


pkg/operator/gcwatchercontroller/gcwatcher_controller.go
```diff
@@ -135,6 +156,19 @@ func (c *GarbageCollectorWatcherController) sync(ctx context.Context, syncCtx fa
...
 	}
+	progressingMonitoringCond := clusteroperatorhelpers.FindStatusCondition(monitoringClusterOperator.Status.Conditions, "Progressing")
+	// If we just started cluster monitoring stack rollout
+	// Time-out after one hour in case cluster monitoring operator gets stuck, so we can report degraded KCM
+	if (progressingMonitoringCond != nil &&
+		progressingMonitoringCond.Status == configv1.ConditionTrue &&
+		progressingMonitoringCond.LastTransitionTime.After(time.Now().Add(-monitoringStackDeployTimeout))) ||
+		(progressingMonitoringCond == nil && monitoringClusterOperator.CreationTimestamp.After(time.Now().Add(-monitoringStackDeployTimeout))) {
+		// To prevent degradation of KCM when installing the cluster monitoring stack or when a new version of cluster monitoring is being rolled out
+		klog.V(5).Info("Monitoring is being rolled out in the cluster and a diagnostic of the garbage collector is not available at this moment. Please look at the kcm logs for more information to debug the garbage collector further")
```
wking, line 167: nit: "the kcm logs" may be enough for folks who spend time in kube-controller-manager space.  And this line will show up in the operator,  so maybe even folks who aren't aware of the acronym can figure it out.  But it would be nice if this reference could be more explicit.  Something like:

> Please see 'oc -n openshift-kube-controller-manager logs -l app=kube-controller-manager --tail -1' for more information to debug the garbage collector.

atiratree: I have updated the message to expand the shortcut to `kube-controller-manager`. I would not prefer to add more detailed debugging information since it is hard coding the deployment of KCM. Also, somebody who is inspecting the KCM operator logs, probably knows  how to inspect KCM logs as well.

pkg/operator/gcwatchercontroller/prometheus_client.go
```diff
@@ -37,12 +37,14 @@ func newPrometheusClient(ctx context.Context, configMapClient corev1client.Confi
 	roots := x509.NewCertPool()
 	roots.AppendCertsFromPEM(bundlePEM)
 
+	dialer := &net.Dialer{
```
ingvagabund, line 40: nit: Why do you need a new variable here? It is used only once and for read-only op

ingvagabund:
@eranco74 can you please validate the PR?

pkg/operator/gcwatchercontroller/prometheus_client.go
atiratree: this was hitting my IDE linter and I forgot why, and I am not seem to able to trigger it again (maybe due to an update?). I have reverthed this change

atiratree:
jobs were failing due to https://github.com/openshift/release/pull/37909 - should be ok now
/retest-required

PR #708, opened 2023-03-09
Reset library-go to openshift fork

JoelSpeed:
After the promotion of Azure and AWS to out of tree, we had to use a fork to make sure merges happened in the right timing. This resets the go.mod back to the master branch of library-go.


damdo (commented):
/lgtm

PR #711, opened 2023-03-20
[4.13] OCPBUGS-8710: Move PSa back to logging

stlaz:
Bumps the o/API with updated featuresets

/cc @deads2k @ibihim


go.mod
```diff
@@ -23,6 +23,8 @@ require (
 	k8s.io/klog/v2 v2.80.1
 )
 
+replace github.com/openshift/api => github.com/stlaz/api v0.0.0-20230316150409-7dfb8e061a12
```
ibihim, line 26: Is this intended to stay within the PR?

stlaz: it's a fake bump, we can't merge until the API PR merges (which it did - I need to update the PR)

stlaz:
/retest-required
I wonder if there are more things still broken in the build payloads

PR #716, opened 2023-03-27
OCPBUGS-7785: migrate to using lease objects for leader election

soltysh:
Manually picked of https://github.com/openshift/cluster-kube-controller-manager-operator/pull/715
/assign @tkashem


soltysh:
I'm overriding the valid-bug, since this will be critical for 4.13->4.14 upgrade.

PR #719, opened 2023-04-02
change kcm connect to localhost apiserver

lance5890:
[issue 720](https://github.com/openshift/cluster-kube-controller-manager-operator/issues/720)
1. there is no need for kcm connect apiserver via APIServerInternalURL
2. change the kcm connect to apiserver via localhost apiserver


lance5890:
@ingvagabund

PR #721, opened 2023-04-04
OCPBUGS-11352: AWS should not use external-cloud-volume-plugin post CSI migration

JoelSpeed:
We believe that in 4.14, the external-cloud-volume-plugin is no longer required for AWS. CSI migration went GA in 4.12 so by 4.14 all volumes should have been migrated across and the in-tree mechanism should no longer be required

CC @bertinatto @soltysh @jsafrane


JoelSpeed:
On the CCM jobs, looks like some nodes never got to requesting their CSRs, which I've seen a lot lately, will poke TRT. Aside from that these jobs don't mean much anymore and should be cleaned up

Will retest the upgrade
/retest-required

soltysh (approved):
/approve

bertinatto:
/lgtm

I've created a custom bundle with this patch and KCM started correctly:

```
$ oc get co/kube-controller-manager
NAME                      VERSION                                                    AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
kube-controller-manager   4.14.0-0.ci.test-2023-04-04-140435-ci-op-yctfdf0i-latest   True        False         False      37m
```

PR #731, opened 2023-04-26
[OCPCLOUD-2034] Update cloud provider tests with feature gate changes

JoelSpeed:
This updates the cloud provider and external volume plugin config observers to use the new feature gates accessor interface. This will allow the feature gate status observation to be determined by the cluster config operator in a synchronised way across the multiple interested operators.

Required to get Azure and GCP CCMs to be promotable in a single PR.

/hold until library-go PR is merged

CC @deads2k

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -3,7 +3,6 @@ package render
 import (
 	"errors"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"path/filepath"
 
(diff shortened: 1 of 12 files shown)


deads2k:
/approve

lgtm too, but needs real bump

JoelSpeed:
I think this can be improved with some changes coming in library-go, but I want to see if the E2E is working now the installer changes are through

pkg/cmd/render/render.go
```diff
@@ -45,17 +45,27 @@ func NewRenderCommand(errOut io.Writer) *cobra.Command {
 		Use:   "render",
 		Short: "Render kubernetes controller manager bootstrap manifests, secrets and configMaps",
 		Run: func(cmd *cobra.Command, args []string) {
-			must := func(fn func() error) {
+			must := func(fn func() error) error {
```
deads2k, line 48: the whole point of `must` is that it doesn't return an error.

JoelSpeed: Yes, I know, but it makes it un-unit-testable.

Normally the code flow would call `klog.Fatalf`, but that would kill the test. But it means the unit tests cannot be representative when an error comes out of any of the three stages of `must`.

In this case, one of the tests checks what happens when there are multiple, contradictory feature gates, the validate returns an error, but the rest of the steps continue uninterrupted, producing an incorrect output

deads2k:
/override ci/prow/e2e-aws-ovn-upgrade 
/override ci/prow/e2e-aws-ovn

unrelated failures, and I need the unit testing content for the next step.

PR #732, opened 2023-04-28
Restricted featuregate reader

deads2k:
Update library-go to get a more restrictive featuregate reader to avoid naughty usage.


pkg/operator/configobservation/cloud/observe_cloud_volume_plugin.go
```diff
@@ -10,10 +11,20 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 )
 
+func NewOObserveCloudVolumePluginFunc(featureGateAccessor featuregates.FeatureGateAccess) configobserver.ObserveConfigFunc {
```
benluddy, line 14:
```suggestion
func NewObserveCloudVolumePluginFunc(featureGateAccessor featuregates.FeatureGateAccess) configobserver.ObserveConfigFunc {
```

deads2k:
external ccm bug

benluddy:
/lgtm
/hold

Holding for a workloads team review.

deads2k:
/hold cancel

if there is a problem, we can revert, though library-go will eventually need udpating.

PR #735, opened 2023-05-10
Remove featureset flag and use only the manifest

deads2k:
The manifest now exists in every install and can be used by bootstrapping. The installer passes the arg now, so we can update the operator to only honor the manifest and eliminate interpretation.

peer to https://github.com/openshift/cluster-kube-apiserver-operator/pull/1491


ingvagabund:
@deads2k is there a jira card that describes the work?

deads2k:
Came through as a bug to unrevert the external CCM feature.  

last cleanup for command related to external CCM feature unrevert. Description of overall flow here: https://github.com/openshift/enhancements/pull/1373

PR #738, opened 2023-05-23
OCPBUGS-13579: remove RBAC for cluster-policy-controller CM leader election

atiratree:
after a recent kube & library-go update https://github.com/openshift/cluster-policy-controller/pull/113 we no longer need RBAC for configmaps in cluster-policy-controller as we have switched to leases exclusively

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -58,6 +58,7 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 		operatorclient.OperatorNamespace,
 		operatorclient.TargetNamespace,
 		"kube-system",
+		"openshift-infra",
 	)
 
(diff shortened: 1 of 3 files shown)


pkg/operator/starter.go
```diff
@@ -58,6 +58,7 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 		operatorclient.OperatorNamespace,
 		operatorclient.TargetNamespace,
 		"kube-system",
+		"openshift-infra",
```
ingvagabund, line 61: Which resources are getting watched in this namespace? Anything new popped up in this namespace? Or, has it been forgotten?

ingvagabund:
Don't see any conflict with https://github.com/openshift/hypershift/tree/main/control-plane-operator/controllers/hostedcontrolplane/kcm.
/lgtm

PR #746, opened 2023-09-15
Bump openshift/* libs

bertinatto:
/assign @soltysh


bertinatto:
/test all
504s everywhere

PR #769, opened 2023-11-22
OCPBUGS-24215: Annotate rotated certificates

vrutkovs:


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -707,6 +708,8 @@ func manageServiceAccountCABundle(ctx context.Context, lister corev1listers.Conf
 	requiredConfigMap, err := resourcesynccontroller.CombineCABundleConfigMaps(
 		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.TargetNamespace, Name: "serviceaccount-ca"},
 		lister,
+		"apiserver-auth",
```
deads2k, line 711: this doesn't look right.  The producer and consumer here is the kube-controller-manager.  I don't think this can simply be assigned to the auth team.  There is *some* claim that the kube-apiserver team owns connections to the kube-apiserver and I'd let that play out between @ingvagabund and @tkashem.

cc @stlaz

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -723,6 +726,8 @@ func ManageCSRCABundle(ctx context.Context, lister corev1listers.ConfigMapLister
 	requiredConfigMap, err := resourcesynccontroller.CombineCABundleConfigMaps(
 		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.OperatorNamespace, Name: "csr-controller-ca"},
 		lister,
+		"apiserver-auth",
```
deads2k, line 729: this doesn't look right. The producer here is the kube-controller-manager and the consumer is kube-apiserver. I don't think this can simply be assigned to the auth team. There is some claim that the kube-apiserver team owns authentication connections for client certificates signed by the kube-controller-manager for the kube-apiserver, but it feels a stretch.

cc @stlaz

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -839,8 +848,14 @@ func ManageCSRIntermediateCABundle(ctx context.Context, lister corev1listers.Sec
 	csrSignerCA, err := client.ConfigMaps(operatorclient.OperatorNamespace).Get(ctx, "csr-signer-ca", metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
 		csrSignerCA = &corev1.ConfigMap{
-			ObjectMeta: metav1.ObjectMeta{Namespace: operatorclient.OperatorNamespace, Name: "csr-signer-ca"},
-			Data:       map[string]string{},
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: operatorclient.OperatorNamespace,
+				Name:      "csr-signer-ca",
+				Annotations: map[string]string{
+					annotations.OpenShiftComponent: "apiserver-auth",
```
deads2k, line 855: this looks solidly owned by kube-controller-manager.  The signer is being run in that process and this operator manages its lifecycle.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
vrutkovs: Updated

PR #778, opened 2023-12-14
OCPBUGS-5825: Removes legacy cloud-provider resources

theobarberbany:
This change removes the cloud-provider ClusterRole, ClusterRoleBinding and ServiceAccount that are no longer required from 4.16

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -138,8 +138,6 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 			"assets/kube-controller-manager/localhost-recovery-token.yaml",
 			"assets/kube-controller-manager/csr_approver_clusterrole.yaml",
 			"assets/kube-controller-manager/csr_approver_clusterrolebinding.yaml",
-			"assets/kube-controller-manager/gce/cloud-provider-role.yaml",
-			"assets/kube-controller-manager/gce/cloud-provider-binding.yaml",
 		},
 		(&resourceapply.ClientHolder{}).WithKubernetes(kubeClient),
 		operatorClient,
@@ -165,6 +163,21 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 			return isVSphere
 		},
 		nil,
+	).WithConditionalResources(
+		bindata.Asset,
+		[]string{
+			"assets/kube-controller-manager/gce/cloud-provider-role.yaml",
+			"assets/kube-controller-manager/gce/cloud-provider-binding.yaml",
+		},
+		func() bool {
+			// We do not want to apply these resources, so must return false here
+			return false
+		},
+		func() bool {
+			// The resources above are required for the 4.14 -> 4.15 upgrade path.
+			// They are not required from 4.16, so can re removed.
+			return true
+		},
 	).AddKubeInformers(kubeInformersForNamespaces)
 
 	targetConfigController := targetconfigcontroller.NewTargetConfigController(


JoelSpeed:
Code change looks ok, should verify that both install and upgrade jobs do not have these resources in their must-gathers once the CI runs complete

pkg/operator/starter.go
```diff
@@ -165,6 +165,22 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 			return isVSphere
 		},
 		nil,
+	).WithConditionalResources(
+		bindata.Asset,
+		[]string{
+			"assets/kube-controller-manager/gce/cloud-provider-role.yaml",
+			"assets/kube-controller-manager/gce/cloud-provider-binding.yaml",
+			"assets/kube-controller-manager/sa.yaml",
```
atiratree, line 173: this sa should be kept. I think the sa for cloud-provider comes from kube

pkg/operator/starter.go
```diff
@@ -165,6 +165,22 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 			return isVSphere
 		},
 		nil,
+	).WithConditionalResources(
+		bindata.Asset,
+		[]string{
+			"assets/kube-controller-manager/gce/cloud-provider-role.yaml",
```
atiratree, line 171: please remove these two files from the main list

pkg/operator/starter.go
```diff
@@ -165,6 +165,22 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
...
+			"assets/kube-controller-manager/gce/cloud-provider-binding.yaml",
+			"assets/kube-controller-manager/sa.yaml",
+		},
+		func() bool {
+			// We do not want to apply these resources, so must return false here
+			return false
+		},
+		func() bool {
+			// The resources above are required for the 4.14 -> 4.15 upgrade path.
```
atiratree, line 180: are you planning to backport this to 4.15 branch?

atiratree (changes requested):
@theobarberbany thx for the PR, but there are a few changes required

pkg/operator/starter.go
JoelSpeed: No, this would break 4.14 to 4.15 upgrades, that's why we are doing this now. The file needs to be removed only once the upgrade to 4.15 is complete, it cannot be removed mid update as that could cause KCM to fall over and halt the upgrade.

pkg/operator/starter.go
ingvagabund: I missed this one. Thanks for noticing it.

pkg/operator/starter.go
atiratree: I mean, 4.15 is not released yet, so I think it could still work if we do it before the release.

atiratree: ^ if the RBAC is not used in the 4.14

pkg/operator/starter.go
theobarberbany: Thanks for the review @atiratree  :) 

> I think the sa for cloud-provider comes from kube

Does this mean there's a SA somewhere else I need to remove?

pkg/operator/starter.go
JoelSpeed: I've been working on the basis that RBAC is used in 4.14, that's why we are removing now, but I will double check when we stopped using it.

JoelSpeed: Yep, confirmed, RBAC is used in 4.14, not used in 4.15, so I want to remove only from 4.16 to avoid breaking upgrades

JoelSpeed:
I'm good with this if @atiratree is, but note, do not want to backport to avoid potential breaking of upgrades

pkg/operator/starter.go
atiratree: I think the SA should be created automatically when the controller runs

https://github.com/kubernetes/kubernetes/blob/a3adc759a3c096accafb1694f77362ee5f957a22/staging/src/k8s.io/legacy-cloud-providers/gce/gce.go#L652

so, no action required

pkg/operator/starter.go
atiratree: Thanks for looking into this. Ok let's remove this in 4.16 then.

pkg/operator/starter.go
theobarberbany: thanks! :)

PR #794, opened 2024-02-28
OCPCLOUD-2514: Remove reliance on Cloud Provider feature gates for determining if CCM should be deployed.

JoelSpeed:
Library go no longer requires feature gates for observing cloud provider state. This PR updates library go to remove the reliance, and ensures that feature gate access is no longer being passed into the cloud provider observers.

A follow up to this is needed to make sure that the `--cloud-provider`, `--external-cloud-volume-plugin` and `--cloud-config` flags are cleared now they are no longer required, will add that as a separate PR later.


elmiko:
i don't think the latest failures are related, but i'm not that familiar with them.

/approve
/lgtm

/test e2e-aws-ovn

PR #795, opened 2024-02-29
OCPCLOUD-2515: remove CloudControllerOwner condition

elmiko:
This change replaces the current logic to set the CloudControllerOwner condition to instead remove the condition whenever it is seen on the ClusterOperator. This is being done because all cloud controllers are now external and cannot be owned by the KCMO.

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -34,7 +34,6 @@ import (
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/crypto"
 	"github.com/openshift/library-go/pkg/operator/certrotation"
-	"github.com/openshift/library-go/pkg/operator/condition"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/management"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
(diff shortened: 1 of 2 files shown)


elmiko:
i would have liked to have a unit test for this, but it seems like we don't have a good fixture for testing the `sync` function.

JoelSpeed:
/lgtm

This is looking good per my expectation from the task, thanks @elmiko

elmiko:
rebased

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -323,68 +322,35 @@ func createTargetConfigController(ctx context.Context, syncCtx factory.SyncConte
...
 
-// setCloudControllerOwnerCondition sets the condition to False if either external cloud
-// provider has been successfully applied for all static pods or it's not set at all. Otherwise
-// it sets the condition to True.
-func setCloudControllerOwnerCondition(ctx context.Context, c TargetConfigController) error {
+// clearCloudControllerOwnerCondition removes the CloudControllerOwner condition if it exists.
+// Prior to version 4.15 of OpenShift, this condition was used to signal the ownership of the
+// cloud controllers. After 4.15 this condition is no longer needed as the external cloud
+// controllers are now the default.
+func removeCloudControllerOwnerCondition(ctx context.Context, c TargetConfigController) error {
```
atiratree, line 329: Can you please add a TODO here that this should be removed in 4.17?

Also, there seems to be other cloud provider bits that should be removed? Is this planned?

atiratree: oh I see this is being worked on in https://github.com/openshift/cluster-kube-controller-manager-operator/pull/799

elmiko: @atiratree did you still want me to ad a TODO here?

atiratree: yes please

atiratree:
thx
/lgtm

PR #801, opened 2024-04-09
OCPBUGS-31807: use RotatedSigningCASecret and RotatedSelfSignedCertKeySecret only in update mode

p0lyn0mial:
UseSecretUpdateOnly is intended as a short term hack for a very specific use case,
and it works in tandem with a particular carry patch applied to the openshift kube-apiserver.
(https://github.com/openshift/kubernetes/pull/1924)

we will remove this when we migrate all of the affected secret
objects to their intended type: https://issues.redhat.com/browse/API-1800

in short tls secrets used by this operator are reconciled
by multiple controllers ([the cert controller run by the operator](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/pkg/operator/certrotationcontroller/certrotationcontroller.go#L85) and [the same controller run by the recovery pod](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/pkg/cmd/recoverycontroller/cmd.go#L92)) at the same time without any coordination.

the issue is that the secret's crypto material
can be regenerated, which has serious consequences for the platform
as it can break external clients and the cluster itself.

xref: https://github.com/openshift/library-go/pull/1710
xref: https://github.com/openshift/kubernetes/pull/1939

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/openshift/api v0.0.0-20231129134630-a782d1c1541c
 	github.com/openshift/build-machinery-go v0.0.0-20230824093055-6a18da01283c
 	github.com/openshift/client-go v0.0.0-20231005121823-e81400b97c46
-	github.com/openshift/library-go v0.0.0-20240403193518-3e3d5f9cf7a4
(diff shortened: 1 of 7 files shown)


p0lyn0mial:
/hold

until https://github.com/openshift/kubernetes/pull/1939 merges

PR #802, opened 2024-04-10
WRKLDS-1169: tolerate node-role.kubernetes.io/control-plane:NoExecute

knelasevero:
Required to be able to advise customers to follow workflow described in https://github.com/openshift/enhancements/pull/1583. If they want to taint control-plane nodes to avoid any other thing to run there (while enforcing it further with validating admission policies), control plane workloads need to tolerate that taint.


knelasevero:
The actual operand doesn't need it as it is static pod and it has Operator:Exists Toleration

PR #810, opened 2024-05-28
Update PodDisruptionBudgetAtLimit alert

sradco:
Update the PodDisruptionBudgetAtLimit  alert to not fire for
kubevirt-disruption-budge, since they are not needed in this use case.

This PR is very important since the alerts fire for every namespace that is running VM, even tough it is not needed and is causing an alert fatigue.

Jira-url: https://issues.redhat.com/browse/CNV-33834

Signed-off-by: Shirly Radco <sradco@redhat.com>

### DIFF
--- a/manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
+++ b/manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
@@ -28,7 +28,7 @@ spec:
             description: The pod disruption budget is at the minimum disruptions allowed level. The number of current healthy pods is equal to the desired healthy pods.
             runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-kube-controller-manager-operator/PodDisruptionBudgetAtLimit.md
           expr: |
-            max by(namespace, poddisruptionbudget) (kube_poddisruptionbudget_status_current_healthy == kube_poddisruptionbudget_status_desired_healthy and on (namespace, poddisruptionbudget) kube_poddisruptionbudget_status_expected_pods > 0)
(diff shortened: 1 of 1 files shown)


ingvagabund:
@sradco is there any cards tracking this change?

sradco:
/assign ingvagabund 
Please review.
This is a high priority issue for OpenShift Virtualization that is causing a lot of noise for our users.

ingvagabund:
Is there a customer case? A bug report?

sradco:
> Is there a customer case? A bug report?

Yes. updated also in the description now. https://issues.redhat.com/browse/CNV-33834.

atiratree:
This has been also suggested in https://github.com/openshift/cluster-kube-controller-manager-operator/pull/583# , but was ultimately [discarded](https://github.com/openshift/cluster-kube-controller-manager-operator/pull/583#issuecomment-1429853464) in favor of silencing.

There is a need for more generic and complete approach. Alternative solution to eviction is being discussed in the Evacuation API KEP upstream.

sradco:
@atiratree Hi, This doesn't align with the effort to reduce alerts noise.
We have an alert to address this issue with a dedicated runbook that explain how this should be addressed in the VMs use case, https://github.com/kubevirt/monitoring/blob/main/docs/runbooks/VMCannotBeEvicted.md.

Silencing requires an action by the user, and the `PodDisruptionBudgetAtLimit` alert only adds confusion in this case, since it will not recommend the correct action, https://github.com/openshift/runbooks/blob/master/alerts/cluster-kube-controller-manager-operator/PodDisruptionBudgetAtLimit.md#mitigation.

In our case the admin needs to set the `evictionStrategy` of the VMI to `shutdown`.

pgodowski:
@sradco our customers are running into a very similar issue with the alert being triggered for the PDB at limit, for the cloud-native-pg (postgres) scenario, where there is PDB create for a primary postgres replica.

The update https://github.com/openshift/cluster-kube-controller-manager-operator/pull/810#issuecomment-2191332099 says that the kubevirt situation is to be handled using silences API - any references what exactly was done for kubevirt, so that possible the same can be evaluated for other use cases?

PR #824, opened 2024-11-19
API-1835: bump library-go

p0lyn0mial:
to pull all changes that have been made.


deads2k:
Without staying in sync with library-go we will be unable to provide service to 4.18

/lgtm
/approve

PR #828, opened 2025-01-17
NO-ISSUE: Bump TraceAll log level to v=10

lance5890:
- and the kube-scheduler-operator has already set the TraceAll to 10, as seen in https://github.com/openshift/cluster-kube-scheduler-operator/blame/b1cc4471e2f6c5dc81b2b9471f4634f1ecdb88b4/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L267

- and the microshift has also bump this TraceAll loglevel to 10  https://github.com/openshift/microshift/pull/4400

If we use TraceALL, we want to trace all the logs really, and some important k8s logs are v(10)


atiratree:
seems reasonable, `persistentvolume-attach-detach-controller` uses `V(10)` logging
/approve
/ok-to-test

lance5890:
> seems reasonable, `persistentvolume-attach-detach-controller` uses `V(10)` logging /approve /ok-to-test

should we move forward ?

PR #834, opened 2025-02-25
STOR-2267: Run SELinux warning controller

jsafrane:
selinux-warning-controller is an optional controller in KCM that emits metrics + events about SELinux usage of persistent volumes in the cluster.

Since most Kubernetes distros don't care about SELinux, this controller needs explicit opt-in on KCM cmdline.
OCP needs the controller:
- to explain to users why their pods may not be running.
- to collect metrics about such promebatic pods, to emit alerts and telemetry.

The controller is disabled by default in Kubernetes 1.32 (under SELinuxChangePolicy feature gate), which is available under `DevPreviewNoUpgrade` since [yesterday](https://github.com/openshift/api/pull/2212). It might reach TechPreviewNoUpgrade in 4.19 if everything goes smooth.

Upstream enhancement: https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1710-selinux-relabeling

~WIP: manual testing~

Tested manually with DevPreviewNoUpgrade cluster.


bindata/assets/config/defaultconfig.yaml
```diff
@@ -26,6 +26,7 @@ extendedArguments:
   - "-ttl" # TODO: this is excluded in kube-core, but not in #21092
   - "-bootstrapsigner"
   - "-tokencleaner"
+  - "selinux-warning-controller" # Most upstream clusters don't care about SELinux, therefore this controller is opt-in. OCP cares.
```
atiratree, line 29: @jsafrane Are you willing to add yourself/your team as the owner of this option/controller?

jsafrane: Added some comment, now it's multi-line to keep it nice. PTAL

atiratree: Thanks

PR #835, opened 2025-02-27
NO-JIRA: readme: describe mapping between operatorLogLevel and log level

lance5890:
related to https://github.com/openshift/cluster-kube-controller-manager-operator/pull/833

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,42 @@ Currently the log levels correspond to:
 | Trace    | 6         |
 | TraceAll | 10        |
 
+
+The log level of cluster-kube-controller-manager-operator can be increased by setting `.spec.operatorLogLevel` field:
+```
+$ oc explain KubeControllerManager.spec.operatorLogLevel
+GROUP:      operator.openshift.io
+KIND:       KubeControllerManager
+VERSION:    v1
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -78,6 +78,42 @@ Currently the log levels correspond to:
...
+KIND:       KubeControllerManager
+VERSION:    v1
+FIELD: operatorLogLevel <string>
+DESCRIPTION:
+    operatorLogLevel is an intent based logging for the operator itself.  It
+    does not give fine grained control, but it is a simple way to manage coarse
+    grained logging choices that operators have to interpret for themselves. 
+     Valid values are: "Normal", "Debug", "Trace", "TraceAll". Defaults to
+    "Normal".
+```
```
atiratree, line 95:
```suggestion
Similarly, the log level of cluster-kube-controller-manager-operator can be increased by setting the `.spec.operatorLogLevel` field:
```

- we do not have to duplicate the description here
- nit: missing `the` here and in the `.spec.logLevel field`

lance5890: updated !

atiratree: thanks!

PR #837, opened 2025-04-11
manifests/0000_90_kube-controller-manager-operator_05_alerts: Template console links in alert descriptions

wking:
[Prometheus alerts support Go templating][1], and this pull uses that to provide more context like "which namespace?", "which PodDisruptionBudget?", "where can I find that PDB in the in-cluster web console?", and "what `oc` command would I run to see garbage-collection sync logs?".  This should make understanding the context of the alert more straightforward, with the responder having to dip into labels and guess.

[1]: https://prometheus.io/docs/prometheus/latest/configuration/template_reference/


manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
```diff
@@ -25,7 +25,8 @@ spec:
         - alert: PodDisruptionBudgetAtLimit
           annotations:
             summary: The pod disruption budget is preventing further disruption to pods.
-            description: The pod disruption budget is at the minimum disruptions allowed level. The number of current healthy pods is equal to the desired healthy pods.
+            description: |-
+              The {{ $labels.poddisruptionbudget }} pod disruption budget in the {{ $labels.namespace}} namespace is at the maximum allowed disruption. The number of current healthy pods is equal to the desired healthy pods.{{ with $console_url := "console_url" | query }}{{ if ne (len (label "url" (first $console_url))) 0}} For more information refer to {{ label "url" (first $console_url) }}/k8s/ns/{{ $labels.namespace }}/poddisruptionbudgets/{{ $labels.poddisruptionbudget }}{{ end }}{{ end }}
```
wking, line 29: Hmm, not sure what to do with [the unit failure][1]:

```
: TestYamlCorrectness expand_less	0s
{=== RUN   TestYamlCorrectness
    assets_test.go:2 ...  === RUN   TestYamlCorrectness
    assets_test.go:2 ...}
```

the test-case's stdout includes:

```
=== RUN   TestYamlCorrectness
    assets_test.go:20: Unexpected error reading manifests from ../../manifests/: failed to render "0000_90_kube-controller-manager-operator_05_alerts.yaml": template: 0000_90_kube-controller-manager-operator_05_alerts.yaml:29: undefined variable "$labels"
```

I guess the that's [this `assets.New`][2] call [through][3] [`assetFromTemplate`][4] through `renderFile` to [this `template.New`][5].  I'm not clear on why this operator feels like these manifests should be Go templates.  Maybe we can pivot to using [`ManifestsFromFiles`][6].

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-kube-controller-manager-operator/837/pull-ci-openshift-cluster-kube-controller-manager-operator-master-unit/1910783714867023872
[2]: https://github.com/openshift/cluster-kube-controller-manager-operator/blob/75f30cdb07d58ce71c0a247207a5d91799a48cc0/pkg/test/assets_test.go#L18C20-L20
[3]: https://github.com/openshift/cluster-kube-controller-manager-operator/blob/75f30cdb07d58ce71c0a247207a5d91799a48cc0/vendor/github.com/openshift/library-go/pkg/assets/assets.go#L43-L45
[4]: https://github.com/openshift/cluster-kube-controller-manager-operator/blob/75f30cdb07d58ce71c0a247207a5d91799a48cc0/vendor/github.com/openshift/library-go/pkg/assets/assets.go#L113
[5]: https://github.com/openshift/cluster-kube-controller-manager-operator/blob/75f30cdb07d58ce71c0a247207a5d91799a48cc0/vendor/github.com/openshift/library-go/pkg/assets/template.go#L68-L69
[6]: https://github.com/openshift/library-go/blob/ce2ba53fb2a4b7318769480e897a0e506b7cb0c0/pkg/manifest/manifest.go#L342-L347

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
```diff
@@ -25,7 +25,8 @@ spec:
         - alert: PodDisruptionBudgetAtLimit
```
wking, line 25: New thread for Cluster Bot testing.  As of daae2166273f0ae, with a `launch 4.19,openshift/cluster-kube-controller-manager-operator#837 aws` cluster, make a PDB mad:

```console
$ oc adm cordon -l node-role.kubernetes.io/worker=
$ oc -n openshift-monitoring delete pod prometheus-k8s-0
$ oc -n openshift-monitoring get poddisruptionbudget prometheus-k8s
NAME             MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
prometheus-k8s   1               N/A               0                     42m
```

I didn't wait for the alert to kick over into `firing`, but checking on `pending`, this looks... almost good to me:

image

the issue is the `<span class="co-resource-item monitoring__resource-item--monitoring-alert co-resource-item--inline">` bit for the `NS` injected into my attempt at constructing a console link.

To trip `PodDisruptionBudgetLimit` I'll look to a different workload, since I don't want to completely break Prometheus (it would make it hard to test alert behavior):

```console
$ oc adm cordon -l node-role.kubernetes.io/master=
$ oc -n openshift-console delete -l component=downloads pods
$ oc -n openshift-console get poddisruptionbudget downloads
NAME        MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
downloads   N/A             1                 0                     53m
```

In that case, the rendering looks great, although I'm not clear on why it's not seeing the `NS` rendering issue:

image

I'm also not clear on how to trigger `GarbageCollectorSyncFailed` to test its rendering.

wking: daae2166273f0a -> 9331433c28f8eb3 added some whitespace before a `}}` to try to get closer to what the working `PodDisruptionBudgetLimit` `description` is doing:

```console
$ git diff --word-diff daae2166273f0a..9331433c28f8eb3 -U0
diff --git a/manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml b/manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
index 8135439..5f299b9 100644
--- a/manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
+++ b/manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
@@ -29 +29 @@ spec:
              The {{ $labels.poddisruptionbudget }} pod disruption budget in the {{ [-$labels.namespace}}-]{+$labels.namespace }}+} namespace is at the maximum allowed disruption. The number of current healthy pods is equal to the desired healthy pods.{{ with $console_url := "console_url" | query }}{{ if ne (len (label "url" (first $console_url))) 0}} For more information refer to {{ label "url" (first $console_url) }}/k8s/ns/{{ $labels.namespace }}/poddisruptionbudgets/{{ $labels.poddisruptionbudget }}{{ end }}{{ end }}
```

But sadly the `NS` markup injected into the middle of the console PDB link is still there:

image

atiratree: Yeah, this is not really optimal. As far as I can see, it should be pretty simple by adding poddisruptionbudget resource to here: https://github.com/openshift/monitoring-plugin/blob/6f948e4323bdf7c68e6b625ce3020116b5b4571a/web/src/components/alerting/AlertsDetailPage.tsx#L450

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
```diff
@@ -25,7 +25,8 @@ spec:
...
           annotations:
             summary: The pod disruption budget is preventing further disruption to pods.
-            description: The pod disruption budget is at the minimum disruptions allowed level. The number of current healthy pods is equal to the desired healthy pods.
+            description: |-
+              The {{ $labels.poddisruptionbudget }} pod disruption budget in the {{ $labels.namespace }} namespace is at the maximum allowed disruption. The number of current healthy pods is equal to the desired healthy pods.{{ with $console_url := "console_url" | query }}{{ if ne (len (label "url" (first $console_url))) 0}} For more information refer to {{ label "url" (first $console_url) }}/k8s/ns/{{ $labels.namespace }}/poddisruptionbudgets/{{ $labels.poddisruptionbudget }}{{ end }}{{ end }}
```
atiratree, line 29: Then we would not need the console url and let the console handle all the link rendering. It would also make the description nicer when looking at the alert detail in `/monitoring/alertrules/1234`.

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
```diff
@@ -35,17 +36,19 @@ spec:
...
+            max by (namespace, poddisruptionbudget) (kube_poddisruptionbudget_status_desired_healthy - kube_poddisruptionbudget_status_current_healthy) > 0
           for: 15m
           labels:
             severity: critical
         - alert: GarbageCollectorSyncFailed
           annotations:
             summary: There was a problem with syncing the resources for garbage collection.
-            description: Garbage Collector had a problem with syncing and monitoring the available resources. Please see KubeControllerManager logs for more details.
+            description: |-
+              Garbage Collector had a problem with syncing and monitoring the available resources. Please see KubeControllerManager logs for more details: 'oc -n {{ $labels.namespace }} logs -c {{ $labels.container }} {{ $labels.pod }}'{{ with $console_url := "console_url" | query }}{{ if ne (len (label "url" (first $console_url))) 0}} For more information refer to {{ label "url" (first $console_url) }}/k8s/ns/{{ $labels.namespace }}/pods/{{ $labels.pod }}/logs?container={{ $labels.container }} {{ end }}{{ end }}.
```
atiratree, line 51: This seems too verbose to me. How to invoke the logs  should be a responsibility of the runbook IMO.

But similar to the PDB case, we could link directly to the pod in the console: https://github.com/openshift/monitoring-plugin/blob/6f948e4323bdf7c68e6b625ce3020116b5b4571a/web/src/components/alerting/AlertsDetailPage.tsx#L450 without too much extra markup.

manifests/0000_90_kube-controller-manager-operator_05_alerts.yaml
atiratree: That is one option.

But we could also try to define the variables and actually try to render it in a test, similar to what we do with other templates https://github.com/openshift/cluster-kube-controller-manager-operator/blob/cec410b7f877a6eef57968889a644b848bdb540a/pkg/cmd/render/render.go#L119

It might be useful to have a test specifically for rendering the alerts to see that it resolves correctly.

atiratree (commented):
There are a few rough edges, but the general idea is very nice!

PR #842, opened 2025-06-21
Revert "CNTRLPLANE-343: certrotationcontroller: use minutes instead of days when FeatureShortCertRotation is enabled"

vrutkovs:
Reverts openshift/cluster-kube-controller-manager-operator#817

Causes `[sig-architecture] platform pods in ns/openshift-kube-controller-manager should not exit an excessive amount of times`


bertinatto (commented):
/lgtm

PR #845, opened 2025-07-01
RFE-4154: add read only filesystem

dusk125:


ingvagabund:
Caution: setting read only file system may cause trouble when a container tries to create a temporary file/directory. It looks like this is not the case given all e2es are passing. Yet,something like https://github.com/openshift/secondary-scheduler-operator/commit/5cddddfcedd96dcb0247798d888f634275804c26#diff-16e5d9a94f028c7269e195d90421ef7653aefc857df2b075504e61cc58e481c6 might be eventually needed.

PR #847, opened 2025-07-23
RFE-4154: Fix readonly filesystem error when kcm trying to mkdir on /etc/kubern…

dusk125:
…etes/kubelet-plugins


ingvagabund:
```
"/etc/kubernetes/kubelet-plugins/volume/exec" # created by machine-config-operator, owned by storage team/hekumar@redhat.com
```
This needs to be allowed to be accessed. Good catch.

PR #848, opened 2025-07-24
OCPBUGS-53290: Missing endpoint slices for open ports the operator uses

aabughosh:
Port 10357 (cluster-policy-controller) was missing service exposure. 

## Problem

The cluster-policy-controller runs on port 10357 within the kube-controller-manager pod but was not exposed through the service. This caused missing endpoint slices for the port that the operator uses.

so the port dont need to be exposed and it need to be internal  with local host


ingvagabund:
`10357` is currently used only for the startup, liveness and readiness probes. Thus, the port does not need to be connected to any service and exposed further than just for the kubelet. I wonder whether the reported issue is a false alarm.

PR #924, opened 2026-04-08
OCPBUGS-84559: CNTRLPLANE-3107:  e2e testing automation Create network network policy test cases

gangwgr:
Adding network policy test cases


gangwgr:
/testwith openshift/cluster-kube-controller-manager-operator/main/e2e-gcp-operator-disruptive https://github.com/openshift/cluster-kube-controller-manager-operator/pull/919

test/e2e/network_policy.go
```diff
@@ -0,0 +1,136 @@
...
+const (
+	kcmOperatorNamespace  = "openshift-kube-controller-manager-operator"
+	kcmTargetNamespace    = "openshift-kube-controller-manager"
+	defaultDenyPolicyName = "default-deny-all"
+	allowOperatorPolicy   = "allow-operator"
+	allowKCMPolicy        = "allow"
+	allowInstallerPolicy  = "allow-installer"
+)
+
+var _ = g.Describe("[sig-api-machinery] kube-controller-manager NetworkPolicy", func() {
```
atiratree, line 51: Why is this for sig-api-machinery? Don't we want to test things specific only to KCM or KCM-o here?

test/e2e/network_policy.go
```diff
@@ -0,0 +1,136 @@
...
+	g.It("[NetworkPolicy][Serial][Disruptive] should ensure KCM NetworkPolicies are defined", func() {
```
atiratree, line 52: Could you specify why do we need these labels?

test/e2e/network_policy.go
```diff
@@ -0,0 +1,136 @@
...
+	installerAllow := getNetworkPolicy(ctx, kubeClient, kcmTargetNamespace, allowInstallerPolicy)
+	logNetworkPolicySummary("[operand] allow-installer", installerAllow)
+	logNetworkPolicyDetails("[operand] allow-installer", installerAllow)
+	requirePodSelectorMatchExpression(installerAllow, "app", metav1.LabelSelectorOpIn, []string{"installer"})
+	requirePolicyTypes(installerAllow, networkingv1.PolicyTypeEgress)
+	requireNoPolicyType(installerAllow, networkingv1.PolicyTypeIngress)
+	requireEgressPort(installerAllow, corev1.ProtocolTCP, 5353)
+	requireEgressPort(installerAllow, corev1.ProtocolUDP, 5353)
+	requireEgressPort(installerAllow, corev1.ProtocolTCP, 8443)
+	logEgressAllowAllPresence(installerAllow)
```
atiratree: the test/check structure seems similar to the ones above? Wouldn't it be easier to parametrize this with a test case list and a single function?

test/e2e/network_policy_helpers.go
```diff
@@ -0,0 +1,634 @@
...
+		}
+	}
+}
+
+func requireIngressPort(policy *networkingv1.NetworkPolicy, protocol corev1.Protocol, port int32) {
+	g.GinkgoHelper()
+	if !hasPortInIngress(policy.Spec.Ingress, protocol, port) {
+		g.Fail(fmt.Sprintf("%s/%s: expected ingress port %s/%d", policy.Namespace, policy.Name, protocol, port))
+	}
+}
```
atiratree: These functions are quite opaque to read and most likely to debug later. Can we bring the errors to the top level? And solve the repetition problem with better scoped test cases?

test/e2e/network_policy.go
```diff
@@ -81,10 +80,9 @@ func testKCMNetworkPolicies() {
 	requireIngressPort(kcmAllow, corev1.ProtocolTCP, 10257)
 	requireIngressPort(kcmAllow, corev1.ProtocolTCP, 10357)
 	requireIngressPort(kcmAllow, corev1.ProtocolTCP, 9443)
+	requireEgressAllowAll(kcmAllow)
 	requireEgressPort(kcmAllow, corev1.ProtocolTCP, 5353)
 	requireEgressPort(kcmAllow, corev1.ProtocolUDP, 5353)
-	requireEgressPort(kcmAllow, corev1.ProtocolTCP, 8443)
```
atiratree: the vendor commit should not contain any code changes

go.mod
```diff
@@ -26,6 +26,11 @@ require (
 	k8s.io/utils v0.0.0-20251002143259-bc988d571ff4
 )
 
+require (
```
atiratree: why is this in its own block?

go.mod
```diff
@@ -125,3 +131,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v6 v6.3.0 // indirect
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
+
+replace github.com/onsi/ginkgo/v2 => github.com/openshift/onsi-ginkgo/v2 v2.6.1-0.20251001123353-fd5b1fb35db1
```
atiratree, line 135: can you please add an expiation above why we need this fork?

test/e2e/network_policy.go
```diff
@@ -0,0 +1,136 @@
...
+	g.GinkgoWriter.Printf("[operand] checking policy %s/%s (source: bindata/assets/, applied by static resource controller)\n", kcmTargetNamespace, allowKCMPolicy)
+	kcmAllow := getNetworkPolicy(ctx, kubeClient, kcmTargetNamespace, allowKCMPolicy)
+	logNetworkPolicySummary("[operand] allow", kcmAllow)
+	logNetworkPolicyDetails("[operand] allow", kcmAllow)
+	requirePodSelectorLabel(kcmAllow, "kube-controller-manager", "true")
+	requireIngressPort(kcmAllow, corev1.ProtocolTCP, 10257)
+	requireIngressPort(kcmAllow, corev1.ProtocolTCP, 10357)
+	requireIngressPort(kcmAllow, corev1.ProtocolTCP, 9443)
+	requireEgressPort(kcmAllow, corev1.ProtocolTCP, 5353)
+	requireEgressPort(kcmAllow, corev1.ProtocolUDP, 5353)
```
atiratree: I guess I am fine with testing the egress DNS port. But do we need to check for the other ingress ports since they are only for documentation purposes?

 What is the purpose of testing it? If we want to test that the documentation is correct, then we should compare it with the source itself.

test/e2e/network_policy.go
```diff
@@ -0,0 +1,136 @@
...
+	g.GinkgoWriter.Printf("[operator] checking policy %s/%s (source: manifests/, applied by CVO)\n", kcmOperatorNamespace, defaultDenyPolicyName)
+	operatorDefaultDeny := getNetworkPolicy(ctx, kubeClient, kcmOperatorNamespace, defaultDenyPolicyName)
+	logNetworkPolicySummary("[operator] default-deny-all", operatorDefaultDeny)
+	logNetworkPolicyDetails("[operator] default-deny-all", operatorDefaultDeny)
+	requireDefaultDenyAll(operatorDefaultDeny)
+
+	g.By("[operator] Validating allow-operator NetworkPolicy in operator namespace (manifests/)")
+	g.GinkgoWriter.Printf("[operator] checking policy %s/%s (source: manifests/, applied by CVO)\n", kcmOperatorNamespace, allowOperatorPolicy)
+	operatorAllow := getNetworkPolicy(ctx, kubeClient, kcmOperatorNamespace, allowOperatorPolicy)
+	logNetworkPolicySummary("[operator] allow-operator", operatorAllow)
```
atiratree: why do we need such agressive logging here? Do we expect any potential failures or regressions here?

test/e2e/network_policy_helpers.go
```diff
@@ -0,0 +1,634 @@
...
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	agnhostImageRef = "registry.k8s.io/e2e-test-images/agnhost:2.45"
+)
+
+// --- NetworkPolicy fetch / assertions ---
+
+func getNetworkPolicy(ctx context.Context, client kubernetes.Interface, namespace, name string) *networkingv1.NetworkPolicy {
```
atiratree: what benefit does it bring to have the call in a function. Too opaque..

test/e2e/network_policy_helpers.go
```diff
@@ -145,7 +149,11 @@ func hasPort(ports []networkingv1.NetworkPolicyPort, protocol corev1.Protocol, p
 		if p.Port == nil || p.Port.IntValue() != int(port) {
 			continue
 		}
-		if p.Protocol == nil || *p.Protocol == protocol {
+		actualProtocol := corev1.ProtocolTCP
```
atiratree: Is this necessary. Isn't the policy taken live from the cluster?

test/e2e/network_policy.go
```diff
@@ -0,0 +1,136 @@
...
+	o.Expect(err).NotTo(o.HaveOccurred())
+	kubeClient, err := kubernetes.NewForConfig(kubeConfig)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	g.By("[operand] Capturing expected NetworkPolicy specs for operand namespace (reconciled by static resource controller)")
+	g.GinkgoWriter.Printf("[operand] reconcile test targets operand namespace %s (policies from bindata/assets/)\n", kcmTargetNamespace)
+	expectedTargetDefaultDeny := getNetworkPolicy(ctx, kubeClient, kcmTargetNamespace, defaultDenyPolicyName)
+	expectedKCMAllow := getNetworkPolicy(ctx, kubeClient, kcmTargetNamespace, allowKCMPolicy)
+	expectedInstallerAllow := getNetworkPolicy(ctx, kubeClient, kcmTargetNamespace, allowInstallerPolicy)
+
```
atiratree: again, this can be parametrized

test/e2e/network_policy_helpers.go
```diff
@@ -0,0 +1,634 @@
...
+
+func restoreNetworkPolicy(ctx context.Context, client kubernetes.Interface, expected *networkingv1.NetworkPolicy) {
+	g.GinkgoHelper()
+	namespace := expected.Namespace
+	name := expected.Name
+	g.GinkgoWriter.Printf("deleting NetworkPolicy %s/%s\n", namespace, name)
+	err := client.NetworkingV1().NetworkPolicies(namespace).Delete(ctx, name, metav1.DeleteOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	err = wait.PollUntilContextTimeout(ctx, 5*time.Second, 10*time.Minute, true, func(ctx context.Context) (bool, error) {
```
atiratree: could we rather use gomega.Eventually?

test/e2e/network_policy_helpers.go
```diff
@@ -0,0 +1,634 @@
...
+		if len(rule.To) == 0 && len(rule.Ports) == 0 {
+			return true
+		}
+	}
+	return false
+}
+
+// --- reconcile helpers ---
+
+func restoreNetworkPolicy(ctx context.Context, client kubernetes.Interface, expected *networkingv1.NetworkPolicy) {
```
atiratree: This can be moved to top level as well.

test/e2e/network_policy_enforcement.go
```diff
@@ -0,0 +1,203 @@
...
+	o.Expect(err).NotTo(o.HaveOccurred())
+	expectConnectivity(ctx, kubeClient, nsName, clientLabels, serverIPs, 8080, false)
+
+	g.By("Adding egress allow and verifying traffic is permitted")
+	_, err = kubeClient.NetworkingV1().NetworkPolicies(nsName).Create(ctx, allowEgressPolicy("allow-egress", nsName, clientLabels, serverLabels, 8080), metav1.CreateOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+	expectConnectivity(ctx, kubeClient, nsName, clientLabels, serverIPs, 8080, true)
+}
+
+func testKCMTargetNamespaceEnforcement() {
```
atiratree: I don't think it makes much sense to create mock pods for KCM. Most of the network policies are for documentation purposes anyway, since we allow all traffic - at least when NetworkPolicy is concerned. But there are other limitations due to the installer.

If we want to test something, I think we could test metrics endpoints both for KCM and KCM-o pods.

And if we want to test that NetworkPolicy is working correctly, then it should live somewhere else.

test/e2e/network_policy.go
gangwgr: in the OTE disruptive and serial suite to avoid interfering with each other or with the operator.
 so it about filtering cases which cases runs in which suite

gangwgr:
@atiratree Updated comments

go.mod
gangwgr: added

test/e2e/network_policy.go
gangwgr: we need events also generated

go.mod
```diff
@@ -125,3 +128,8 @@ require (
 	sigs.k8s.io/structured-merge-diff/v6 v6.3.0 // indirect
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
+
+// The OpenShift fork of ginkgo exposes internal suite APIs required by
+// openshift-tests-extension (oteginkgo.BuildExtensionTestSpecsFromOpenShiftGinkgoSuite).
+// Upstream ginkgo does not export these symbols; see https://github.com/openshift/onsi-ginkgo.
```
atiratree, line 134: Thanks, just one nit:

```suggestion
// Upstream ginkgo does not export these symbols (e.g. `Suite` ); see https://github.com/openshift/onsi-ginkgo.
```

test/e2e/network_policy.go
atiratree: The first test is read only.

The second test shoud only disable the policy for a short amount of time, so it shouldn't affect the operator/cluster.

Also, why do we need `NetworkPolicy` here?

test/e2e/network_policy.go
atiratree: This still applies

test/e2e/network_policy.go
```diff
@@ -0,0 +1,147 @@
...
+	name      string
+	denyAll   bool
+}
+
+var allPolicies = []policyTestCase{
+	{component: "operator", namespace: kcmOperatorNamespace, name: defaultDenyPolicyName, denyAll: true},
+	{component: "operator", namespace: kcmOperatorNamespace, name: allowOperatorPolicy},
+	{component: "operand", namespace: kcmTargetNamespace, name: defaultDenyPolicyName, denyAll: true},
+	{component: "operand", namespace: kcmTargetNamespace, name: allowKCMPolicy},
+	{component: "operand", namespace: kcmTargetNamespace, name: allowInstallerPolicy},
```
atiratree, line 41: This is a duplication of the operandPolicies. Can we just have operatorPolicies and  operandPolicies and combine them when we need it.

atiratree: Or let's just move each of these to their test. Usually there is not much benefit of sharing these between tests.

test/e2e/network_policy.go
```diff
@@ -0,0 +1,147 @@
...
+				"[%s] %s should select all pods", tc.component, tc.name)
+			o.Expect(policy.Spec.PodSelector.MatchExpressions).To(o.BeEmpty(),
+				"[%s] %s should have no match expressions", tc.component, tc.name)
+			o.Expect(policy.Spec.Ingress).To(o.BeEmpty(),
+				"[%s] %s should have no ingress rules", tc.component, tc.name)
+			o.Expect(policy.Spec.Egress).To(o.BeEmpty(),
+				"[%s] %s should have no egress rules", tc.component, tc.name)
+			continue
+		}
+
```
atiratree, line 84: We could also check that ingress ports are not empty if `Ingress` `policyType` is present.

test/e2e/network_policy.go
```diff
@@ -0,0 +1,147 @@
...
+func testKCMNetworkPolicies() {
+	ctx := context.Background()
+	kubeConfig, err := testlib.NewClientConfigForTest()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	kubeClient, err := kubernetes.NewForConfig(kubeConfig)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	for _, tc := range allPolicies {
+		g.By(fmt.Sprintf("[%s] Validating %s/%s", tc.component, tc.namespace, tc.name))
```
atiratree, line 68: It is easier to include the test case name in the list

test/e2e/network_policy.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func testKCMNetworkPolicyReconcile() {
+	ctx := context.Background()
+	kubeConfig, err := testlib.NewClientConfigForTest()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	kubeClient, err := kubernetes.NewForConfig(kubeConfig)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	for _, tc := range operandPolicies {
```
atiratree, line 104: We should start with the test case name first.

test/e2e/network_policy.go
```diff
@@ -0,0 +1,147 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By(fmt.Sprintf("[%s] Deleting %s/%s and waiting for restoration", tc.component, tc.namespace, tc.name))
+		err = kubeClient.NetworkingV1().NetworkPolicies(tc.namespace).Delete(ctx, tc.name, metav1.DeleteOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		o.Eventually(func(eg o.Gomega) {
+			restored, err := kubeClient.NetworkingV1().NetworkPolicies(tc.namespace).Get(ctx, tc.name, metav1.GetOptions{})
+			eg.Expect(err).NotTo(o.HaveOccurred())
+			eg.Expect(equality.Semantic.DeepEqual(original.Spec, restored.Spec)).To(o.BeTrue(),
```
atiratree, line 115: it is better to do the assertion via the `GomegaMatcher` when using `Eventually`

test/e2e/network_policy.go
```diff
@@ -0,0 +1,147 @@
...
+				"spec mismatch after restore of %s/%s", tc.namespace, tc.name)
+		}).WithTimeout(10*time.Minute).WithPolling(5*time.Second).Should(o.Succeed(),
```
atiratree, line 117:
```suggestion
		}).WithTimeout(5*time.Minute).WithPolling(5*time.Second).Should(o.Succeed(),
```

should be enough

test/e2e/network_policy.go
```diff
@@ -0,0 +1,147 @@
...
+		o.Eventually(func(eg o.Gomega) {
+			current, err := kubeClient.NetworkingV1().NetworkPolicies(tc.namespace).Get(ctx, tc.name, metav1.GetOptions{})
+			eg.Expect(err).NotTo(o.HaveOccurred())
+			eg.Expect(equality.Semantic.DeepEqual(original.Spec, current.Spec)).To(o.BeTrue(),
+				"spec mismatch after mutation of %s/%s", tc.namespace, tc.name)
+		}).WithTimeout(10*time.Minute).WithPolling(5*time.Second).Should(o.Succeed(),
+			"[%s] %s/%s should be reconciled after mutation", tc.component, tc.namespace, tc.name)
+	}
+
+	g.By("Logging NetworkPolicy-related events (best-effort)")
```
atiratree, line 134: Why do we need this? Is this only for debugging? The events are collected during the run and available in the CI artifacts? There are also other debug vectors.

Is this a common practice? If yes, can it be done on a framework level?

test/e2e/network_policy.go
```diff
@@ -0,0 +1,147 @@
...
+		patch := []byte(`{"spec":{"podSelector":{"matchLabels":{"np-reconcile":"mutated"}}}}`)
+		_, err = kubeClient.NetworkingV1().NetworkPolicies(tc.namespace).Patch(ctx, tc.name, types.MergePatchType, patch, metav1.PatchOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		o.Eventually(func(eg o.Gomega) {
+			current, err := kubeClient.NetworkingV1().NetworkPolicies(tc.namespace).Get(ctx, tc.name, metav1.GetOptions{})
+			eg.Expect(err).NotTo(o.HaveOccurred())
+			eg.Expect(equality.Semantic.DeepEqual(original.Spec, current.Spec)).To(o.BeTrue(),
+				"spec mismatch after mutation of %s/%s", tc.namespace, tc.name)
+		}).WithTimeout(10*time.Minute).WithPolling(5*time.Second).Should(o.Succeed(),
```
atiratree, line 130:
```suggestion
		}).WithTimeout(5*time.Minute).WithPolling(5*time.Second).Should(o.Succeed(),
```

test/e2e/network_policy.go
```diff
@@ -0,0 +1,147 @@
...
+	{component: "operand", namespace: kcmTargetNamespace, name: defaultDenyPolicyName, denyAll: true},
+	{component: "operand", namespace: kcmTargetNamespace, name: allowKCMPolicy},
+	{component: "operand", namespace: kcmTargetNamespace, name: allowInstallerPolicy},
+}
+
+var _ = g.Describe("[sig-api-machinery] kube-controller-manager NetworkPolicy", func() {
+	g.It("[NetworkPolicy][Serial][Disruptive] should ensure KCM NetworkPolicies are defined", func() {
+		testKCMNetworkPolicies()
+	})
+	g.It("[NetworkPolicy][Serial][Disruptive] should restore KCM NetworkPolicies after delete or mutation [Timeout:30m]", func() {
```
atiratree, line 55: Also, we don't need this big timeout. What is the default timeout?

gangwgr:
https://github.com/openshift/cluster-kube-controller-manager-operator/pull/924#discussion_r3103351413
it is because we have only added one suite job for one case only not want to add new job

test/e2e/network_policy.go
```diff
@@ -0,0 +1,145 @@
...
+		err = kubeClient.NetworkingV1().NetworkPolicies(tc.namespace).Delete(ctx, tc.name, metav1.DeleteOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		o.Eventually(func() (bool, error) {
+			restored, err := kubeClient.NetworkingV1().NetworkPolicies(tc.namespace).Get(ctx, tc.name, metav1.GetOptions{})
+			if err != nil {
+				return false, err
+			}
+			return equality.Semantic.DeepEqual(original.Spec, restored.Spec), nil
+		}).WithTimeout(5*time.Minute).WithPolling(5*time.Second).Should(o.BeTrue(),
```
atiratree: please prefer using `gomega.Equal` and other GomegaMatchers for better messages.

go.mod
```diff
@@ -125,3 +128,8 @@ require (
 	sigs.k8s.io/structured-merge-diff/v6 v6.3.0 // indirect
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
+
+// The OpenShift fork of ginkgo exposes internal suite APIs required by
+// openshift-tests-extension (oteginkgo.BuildExtensionTestSpecsFromOpenShiftGinkgoSuite).
+// Upstream ginkgo does not export these symbols (e.g. `Suite` ); see https://github.com/openshift/onsi-ginkgo.
```
atiratree:
```suggestion
// Upstream ginkgo does not export these symbols (e.g. `Suite`); see https://github.com/openshift/onsi-ginkgo.
```

test/e2e/network_policy.go
atiratree: This still applies. It is also better to do https://github.com/openshift/cluster-kube-controller-manager-operator/pull/924#discussion_r3103512742

test/e2e/network_policy.go
atiratree: Still applies

test/e2e/network_policy.go
atiratree: How is https://github.com/openshift/cluster-kube-controller-manager-operator/pull/924#issuecomment-4277947862 relevant to the previous questions?

Can you link the jobs you mentioned and explain the problem you see?

gangwgr:
the job https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-kube-controller-manager-operator/924/pull-ci-openshift-cluster-kube-controller-manager-operator-main-e2e-gcp-operator-disruptive/2046099003061309440 regarding https://github.com/openshift/cluster-kube-controller-manager-operator/pull/924#discussion_r3110046218

go.mod
gangwgr: updated

gangwgr:
updated https://github.com/openshift/cluster-kube-controller-manager-operator/pull/924#discussion_r3110003503

test/e2e/network_policy.go
```diff
@@ -33,10 +33,10 @@ type policyTestCase struct {
 }
 
 var _ = g.Describe("[sig-api-machinery] kube-controller-manager NetworkPolicy", func() {
-	g.It("should ensure KCM NetworkPolicies are defined [Serial]", func() {
+	g.It("should ensure KCM NetworkPolicies are defined [Serial][Disruptive]", func() {
 		testKCMNetworkPolicies()
 	})
-	g.It("[NetworkPolicy][Serial][Disruptive] should restore KCM NetworkPolicies after delete or mutation", func() {
+	g.It("should restore KCM NetworkPolicies after delete or mutation [Serial][Disruptive]", func() {
```
atiratree, line 39: This change should not reside in the vendor commit.

test/e2e/network_policy.go
```diff
@@ -0,0 +1,144 @@
...
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		g.By(fmt.Sprintf("%s: deleting %s/%s and waiting for restoration", tc.desc, tc.namespace, tc.name))
+		err = kubeClient.NetworkingV1().NetworkPolicies(tc.namespace).Delete(ctx, tc.name, metav1.DeleteOptions{})
+		o.Expect(err).NotTo(o.HaveOccurred())
+
+		o.Eventually(func() (networkingv1.NetworkPolicySpec, error) {
+			restored, err := kubeClient.NetworkingV1().NetworkPolicies(tc.namespace).Get(ctx, tc.name, metav1.GetOptions{})
+			if err != nil {
+				return networkingv1.NetworkPolicySpec{}, err
```
atiratree: nit: this looks better
```suggestion
				return nil, err
```

test/e2e/network_policy.go
atiratree: The second test affects the first test, so it is indeed disruptive, we can use `e2e-gcp-operator-disruptive`  there as you mentioned.

I still don't see why we need disruptive for the first one though, since it is read only. We will have to add parallel test suite eventually, right?

test/e2e/network_policy.go
atiratree: ^

gangwgr:
@atiratree https://github.com/openshift/cluster-kube-controller-manager-operator/pull/924#discussion_r3118058854 
you are right first case it readonly but we have added only one job which is for to run disruptive cases.  for single case we don't want to create new for single readonly cases.  It is not cost effective. In future when we more cases we can create parallel job and then move this cases to that job and remove the tag

atiratree:
If we don't expect a big influx of tests, then this should be done at the test suite level + documented. Each test should have the correct tags.

gangwgr:
> If we don't expect a big influx of tests, then this should be done at the test suite level + documented. Each test should have the correct tags.

Agreed. I've documented this at the suite level in main.go — added a comment on the disruptive suite explaining that all NetworkPolicy tests currently run there because we only maintain one serial job, and that tags should be adjusted when a parallel suite is introduced. We later migrate go standard cases to ote then we will change. If right now removed that tag from readonly case it will not run any job.

test/e2e/network_policy.go
```diff
@@ -0,0 +1,144 @@
...
+
+type policyTestCase struct {
+	desc      string
+	namespace string
+	name      string
+	denyAll   bool
+}
+
+var _ = g.Describe("[sig-api-machinery] kube-controller-manager NetworkPolicy", func() {
+	g.It("should ensure KCM NetworkPolicies are defined [Serial][Disruptive]", func() {
```
atiratree, line 36:
```suggestion
	g.It("should ensure KCM NetworkPolicies are defined", func() {
```

cmd/cluster-kube-controller-manager-operator-tests-ext/main.go
```diff
@@ -1,53 +1,80 @@
...
+	})
+
+	// All NetworkPolicy tests currently run in the disruptive suite because
+	// we only maintain one serial job. When more read-only tests are added,
+	// a parallel suite can be introduced and their tags adjusted accordingly.
+	extension.AddSuite(oteextension.Suite{
+		Name:        "openshift/cluster-kube-controller-manager-operator/operator/disruptive",
+		Parallelism: 1,
+		Qualifiers: []string{
+			`name.contains("[Serial]") || name.contains("[Disruptive]")`,
```
atiratree, line 67:
```suggestion
```

atiratree: or maybe just comment it out, and mention a TODO - once we enable the parallel tests

cmd/cluster-kube-controller-manager-operator-tests-ext/main.go
```diff
@@ -1,53 +1,80 @@
...
 	registry := oteextension.NewRegistry()
 	extension := oteextension.NewExtension("openshift", "payload", "cluster-kube-controller-manager-operator")
 
+	extension.AddSuite(oteextension.Suite{
+		Name:        "openshift/cluster-kube-controller-manager-operator/operator/parallel",
+		Parallelism: 4,
+		Qualifiers: []string{
+			`!name.contains("[Serial]") && !name.contains("[Disruptive]")`,
+		},
+	})
```
atiratree, line 58: we can comment this out as well, until we are ready to enable the parallel tests

test/e2e/network_policy.go
gangwgr: @atiratree do yo want remove this cases from disruptive?

gangwgr: if we remove tag it will not run in any job

cmd/cluster-kube-controller-manager-operator-tests-ext/main.go
gangwgr: why we need to remove this suite? this one needed

cmd/cluster-kube-controller-manager-operator-tests-ext/main.go
gangwgr: added comments, do you want to disable tis suite? 
we can't  disable cases suite disruptive job will be useless

test/e2e/network_policy.go
gangwgr: it we remove tag from cases it will not run any job

cmd/cluster-kube-controller-manager-operator-tests-ext/main.go
gangwgr: added comments for TODO

test/e2e/network_policy.go
atiratree: It should, if we relax the Qualifiers, no?

cmd/cluster-kube-controller-manager-operator-tests-ext/main.go
```diff
@@ -1,53 +1,82 @@
...
-	return cmd
+	return cmd, nil
 }
 
-func prepareOperatorTestsRegistry() *oteextension.Registry {
+func prepareOperatorTestsRegistry() (*oteextension.Registry, error) {
 	registry := oteextension.NewRegistry()
 	extension := oteextension.NewExtension("openshift", "payload", "cluster-kube-controller-manager-operator")
 
+	// TODO: enable the parallel suite once we have enough read-only tests
```
atiratree:
```suggestion
	// TODO: enable the parallel suite once we have enough non-conflicting non-disruptive tests
```

cmd/cluster-kube-controller-manager-operator-tests-ext/main.go
atiratree: do we have a tls suite here?

cmd/cluster-kube-controller-manager-operator-tests-ext/main.go
atiratree: No removal, just relaxing the qualifiers. We can add a TODO that will return the qualifiers back to normal once we enable the parallel suite.

cmd/cluster-kube-controller-manager-operator-tests-ext/main.go
gangwgr: if we have multiple suites then we should create multiple jobs for that
it will be expensive

test/e2e/network_policy.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	ctx := context.Background()
+	kubeConfig, err := testlib.NewClientConfigForTest()
+	o.Expect(err).NotTo(o.HaveOccurred())
+	kubeClient, err := kubernetes.NewForConfig(kubeConfig)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	for _, tc := range policies {
+		original, err := kubeClient.NetworkingV1().NetworkPolicies(tc.namespace).Get(ctx, tc.name, metav1.GetOptions{})
```
atiratree, line 116: printing the name of the tc should be the first thing we do

gangwgr: updated

cmd/cluster-kube-controller-manager-operator-tests-ext/main.go
gangwgr: updated

cmd/cluster-kube-controller-manager-operator-tests-ext/main.go
gangwgr: yes updated

atiratree: I still don't see this resolved

gangwgr: updatd

cmd/cluster-kube-controller-manager-operator-tests-ext/main.go
```diff
@@ -1,53 +1,76 @@
...
+	// 	Name:        "openshift/cluster-kube-controller-manager-operator/operator/parallel",
+	// 	Parallelism: 4,
+	// 	Qualifiers: []string{
+	// 		`!name.contains("[Serial]") && !name.contains("[Disruptive]")`,
+	// 	},
+	// })
+
+	extension.AddSuite(oteextension.Suite{
+		Name:             "openshift/cluster-kube-controller-manager-operator/operator/disruptive",
+		Parallelism:      1,
```
atiratree, line 65: Please add the TODO here as well. We can also comment out the needed Qualifiers here.

test/e2e/network_policy.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+type policyTestCase struct {
+	desc      string
+	namespace string
+	name      string
+	denyAll   bool
+}
+
+var _ = g.Describe("[kube-controller-manager-operator] NetworkPolicy", func() {
```
atiratree: Has there been a consensus that we are going to use labels in such maner? Or do we have examples of our components using them?

gangwgr: same manner we have to use

gangwgr: I don't think there's an established pattern for component-specific labels like this. Should i change it back to [sig-api-machinery] or to follow like
```
grep: brackets ([ ]) not balanced
rgangwar@rgangwar-mac origin % grep -inr "\[sig-kube"
./test/extended/kubevirt/services.go:13:var _ = Describe("[sig-kubevirt] services", func() {
./test/extended/kubevirt/migration.go:19:var _ = Describe("[sig-kubevirt] migration", func() {
./test/extended/apiserver/rollout.go:36:var _ = ginkgo.Describe("[Conformance][Suite:openshift/kube-apiserver/rollout][Jira:\"kube-apiserver\"][sig-kube-apiserver] kube-apiserver", func() {
```
 or

 https://github.com/openshift/cluster-image-registry-operator/pull/1322/changes#diff-c9f93ea47b1258790a582ba541c229251796c331b0940db211a304a500ef0662R21

PR #899, opened 2025-12-04
Testing: replace with lance5890/library-go//remove_BoundServceAccountTokenVolume

lance5890:

PR #941, opened 2026-06-05
update vendor with https://github.com/openshift/library-go/pull/2001

lance5890:
replace https://github.com/openshift/cluster-kube-controller-manager-operator/pull/899

related to https://github.com/openshift/library-go/pull/2001

## Summary by CodeRabbit

* **Chores**
  * Updated OpenShift module dependencies to newer versions and configured a module redirect for enhanced compatibility.


lance5890:
@tchap

tchap:
There was IMO no need to replace the previous testing PR, but doesn't matter.

/ok-to-test

PR #948, opened 2026-07-08
OCPBUGS-38662: Add 10-minute degraded inertia for StaticPodsDegraded

tchap:
Static pods may temporarily report errors during upgrades while pods restart. Use a longer inertia to avoid flapping the ClusterOperator Degraded condition.

## Summary by CodeRabbit

* **Bug Fixes**
  * Expanded degraded-condition handling so inertia now applies to both node-controller and static-pods degraded states.
  * This helps prevent short-lived status changes from triggering unnecessary degradation alerts.
* **Tests**
  * Updated coverage to validate the added degraded-condition duration behavior.


tchap:
The CI fail is unrelated.

/verified by CI
/retest

PR #950, opened 2026-08-06
OCPBUGS-99770: Remove unnecessary namespace informers from kubeInformersForNamespaces

tchap:
Remove the cluster-wide ("") and "openshift-infra" namespace entries from kubeInformersForNamespaces. Both entries exist solely to provide dynamic wiring for the static resource controller via AddKubeInformers — no other controller in the operator ever queries either of them through InformersFor, ConfigMapLister, SecretLister, or PodLister.

The cluster-wide ("") entry gives dynamic wiring to cluster-scoped static resources (ClusterRoles and ClusterRoleBindings). A separate clusterInformers instance already provides the cluster-wide informer used by the static pod builder's node and guard controllers. The "" entry in kubeInformersForNamespaces creates a second, independent SharedInformerFactory that sets up its own watch connections for RBAC resources across all namespaces.

The "openshift-infra" entry gives dynamic wiring to two static resources: the openshift-infra Namespace and the pv-recycler-controller ServiceAccount. No controller reads configmaps, secrets, or pods from this namespace.

AddKubeInformers gracefully handles missing namespace entries — it logs a warning and skips dynamic wiring, falling back to the static resource controller's time-based reconciliation (every 10 minutes via resync). Since these are stable, rarely-changing resources, time-based reconciliation is sufficient.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource monitoring by limiting watches to the configured operator namespaces and `kube-system`.
  * Removed unnecessary monitoring of the empty namespace and `openshift-infra`.

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -58,14 +58,15 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 	}
 	clusterInformers := v1helpers.NewKubeInformersForNamespaces(kubeClient, "")
 	configInformers := configinformers.NewSharedInformerFactory(configClient, 10*time.Minute)
+	// NamespaceAll ("") is intentionally not included — cluster-scoped static resources
+	// (ClusterRoles, ClusterRoleBindings) lose dynamic informer wiring but are still
+	// reconciled by the static resource controller on its 1-minute resync.
 	kubeInformersForNamespaces := v1helpers.NewKubeInformersForNamespaces(kubeClient,
-		"",
 		operatorclient.GlobalUserSpecifiedConfigNamespace,
 		operatorclient.GlobalMachineSpecifiedConfigNamespace,
 		operatorclient.OperatorNamespace,
 		operatorclient.TargetNamespace,
 		"kube-system",
-		"openshift-infra",
 	)
 
 	operatorClient, dynamicInformers, err := genericoperatorclient.NewStaticPodOperatorClient(


pkg/operator/starter.go
```diff
@@ -59,13 +59,11 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 	clusterInformers := v1helpers.NewKubeInformersForNamespaces(kubeClient, "")
 	configInformers := configinformers.NewSharedInformerFactory(configClient, 10*time.Minute)
 	kubeInformersForNamespaces := v1helpers.NewKubeInformersForNamespaces(kubeClient,
-		"",
 		operatorclient.GlobalUserSpecifiedConfigNamespace,
 		operatorclient.GlobalMachineSpecifiedConfigNamespace,
 		operatorclient.OperatorNamespace,
 		operatorclient.TargetNamespace,
 		"kube-system",
-		"openshift-infra",
```
atiratree, line 68: It should be fine to leave this here as we do not expect many watch events from this namespace, but I don't have a strong opinion on this

pkg/operator/starter.go
```diff
@@ -59,13 +59,11 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 	clusterInformers := v1helpers.NewKubeInformersForNamespaces(kubeClient, "")
 	configInformers := configinformers.NewSharedInformerFactory(configClient, 10*time.Minute)
 	kubeInformersForNamespaces := v1helpers.NewKubeInformersForNamespaces(kubeClient,
-		"",
```
atiratree, line 62: let's document this to explicitely ignore NamespaceAll. We will rely on periodic resync which is 1 minute for syncing ClusterRole and ClusterRoleBindings

which should be sufficient

pkg/operator/starter.go
tchap: It's explicitly mentioned in the Jira issue, so if we don't have a strong opinion, I would remove it. But I also don't have a strong opinion 😅

pkg/operator/starter.go
tchap: Added an explaining comment.

atiratree:
Thx!
/lgtm
/approve
/verified by @atiratree and ci

